PR #1, opened 2017-03-03
Changes for LSE-163

timj:
* US Letter not A4
* Remove \req for now (was Gaia req macro)
* Add some additional commands from LSE-163
* Use standard maketitle/author/title commands

### DIFF
--- a/tex/latex/lsst/lsstdoc.cls
+++ b/tex/latex/lsst/lsstdoc.cls
@@ -5,7 +5,7 @@
 \NeedsTeXFormat{LaTeX2e}
 \ProvidesClass{lsstdoc}[2016/12/12 v1.1 LSST document class]
 
-\LoadClass[12pt,a4paper]{article}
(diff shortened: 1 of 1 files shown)


womullan:
I use \req in a statement of work I am writing with KT …

it is handy for properly labled things ..

timj:
Ok. I used a different `\req` in LSE-163 so I'll change the LSE-163 end of things to use a different macro for requirements annotations.

timj:
I've reverted the `\req` change.

womullan:
i have less dependencies I could switch to yours ..

timj:
no. I've made the fix already. Wasn't much. My requirement anchors added margin notes so I made that explicit in the name.

womullan:
these are more example documents ..  I did try to make them all LSST shaped. I would leave the images in this folder but indeed another repo of share images fo the actual dm-docs would be good.

PR #2, opened 2017-03-10
More closely match standard LSST doc layout

timj:
* Use sans serif font.
* Use author/title layout to match documents.
* ChangeRecord block. This breaks compatibility with Gaia.
* Roman page numbers for preamble.
* Header fixes.
* Use blue for sections.

Note this now requires that document page numbering is reverted to arabic in the main document. One option is to move table of contents creation into `\maketitle` command.


womullan:
ok .. one of the reasons of the templates (and the name may be wrong) is as an example of a working document.  I now have the new texmf and  my existing document fails (LDM-294) .. I have no example of a document that works  so it makes it a little difficult to know why \mktitle no longer exists. So perhaps templates should be EXAMPLES and I am happy to put my examples back in there once I figure out how to fix them all... a real pity LSST use sans serif fonts but that is not an dm issue.   Ok So see its maketitle now  and i guess this is what you mentioned about the \title \author but i have no example -- \seDocTitle still exists .. but then maketitel complains \title is not defined  so I am not sure how to proceed ..

timj:
I'll upload an example :-)

timj:
An example now exists and this PR shows what is needed to change from the current master. I intend to add a `\setDocShortTitle` rather than doing it the current way. I don't think I can bring table of contents into `\maketitle` without having to define the change history in the preamble.

womullan:
ok looks good to me now and i can use it. I fixed a small issue with the change recode (mine overflowed then I build yours and the header did not line up). I checked in an example presentation. I see you do not put the PDF of the example . I understand with an auto build that may not be neeed but for the examples ?... remove as you wish before merging - I added thesse to your PR.

timj:
@womullan I've merged this with the addition of a make file that can be used to build the examples. This make file is also used by travis so when you do a pull request we can get feedback straightaway as to whether the class files have broken something.

PR #3, opened 2017-03-14
Add table of contents and change record to make title

timj:
* New class option "toc" to automatically add table of contents.
  This allows \maketitle to put it in the correct place in the document
  and to fix the page number to arabic.
* Add \setDocChangeRecord to allow change record to be defined
  in the preamble. This allows \maketitle to put it in the correct
  place in the document.

Both these changes allow \maketitle to set up the document in a standard
way rather than relying on the author to do it correctly. These changes
do not affect backwards compatibility.

### DIFF
--- a/examples/LDM-nnn.tex
+++ b/examples/LDM-nnn.tex
@@ -1,4 +1,4 @@
-\documentclass[DM,lsstdraft]{lsstdoc}
+\documentclass[DM,lsstdraft,toc]{lsstdoc}
 
 \usepackage[english]{babel}
 \usepackage[utf8x]{inputenc}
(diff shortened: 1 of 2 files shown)


timj:
Since it was bugging me a lot, I've also pushed a couple of tweaks to allow optional short title with `\title` and `\date` to be used instead of `\setDocDate`. These are the final tweaks that I think are needed for sanity.

womullan:
the readme also lack the command
mktexlsr

without which tex will not find these templates ..

timj:
Do we want tex to be able to find these example files? Are you asking for a note in the top level README reminding people to rerun `mktexlsr` when they add a class/style file to the `texmf` directory?

womullan:
It works fine for meOn 15 Mar 2017 9:42 p.m., Tim Jenness <notifications@github.com> wrote:Ok. I understand. I'll add that note. It sounds like you have no other comments so I'll merge this.

PR #5, opened 2017-03-21
DM-9896: Test the bibtex entries are properly formed

timj:
As we add more bibtex entries it seemed to me that we needed to ensure that our bib files didn't contain obvious formatting problems. Whilst editors can do a lot of this, bibtex itself seems to be a very good linter. Warnings from bibtex are not treated as errors (we have many of them).

This PR adds a new test latex file that uses the `lsstdoc.cls` document class and imports all the bibtex references and creates a document. This is added to the standard make file and therefore also tested by Travis.

The Gaia livelink entries are not vetted because they contain too many errors and are not relevant to LSST.

This PR includes some minor fixes to bibtex files required to allow the test document to be built.

### DIFF
--- a/texmf/bibtex/bib/refs.bib
+++ b/texmf/bibtex/bib/refs.bib
@@ -236,7 +236,7 @@ IMAGINARIES IN ALGEBRA },
   livelinkshortnumber =    {ECSS-M-40B }
 }
 
-@LiveLink{ecss:spmdoc,
+@LiveLink{ecss:spmdoc50B,
   author =  { {ESA Publications Division}},
   title =   { {S}pace {P}roject {M}anagent - information/documentation managment },
   institution = { {E}uropean {C}ooperation for {S}pace {S}tandardization },
@@ -391,7 +391,11 @@ requirements},
   author = {{Schneider}, J.},
   title = "{Can the Perturbation of a Stellar Motion in a Triple System Mimic
             a Planet}",
-   crossref={TDUG},
+  booktitle = {ESA SP-576: The Three-Dimensional Universe with Gaia},
+  year = 2005,
+  editor = {{Turon}, C. and {O'Flaherty}, K.~S. and {Perryman}, M.~A.~C.
+            },
+  month = jan,
    pages = {263--266}
 }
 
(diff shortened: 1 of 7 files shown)


texmf/bibtex/bib/refs.bib
```diff
@@ -236,7 +236,7 @@ @LiveLink{ecss:spmdoc
   livelinkshortnumber =    {ECSS-M-40B }
 }
 
-@LiveLink{ecss:spmdoc,
+@LiveLink{ecss:spmdoc50B,
```
timj: This was a duplicate entry. LSST doesn't really care which one stays as `ecss:spmdoc` but I change the latter one to use `50B` suffix in this case.

texmf/bibtex/bib/refs.bib
```diff
@@ -391,7 +391,11 @@ @INPROCEEDINGS{Schneider-2005:a
   author = {{Schneider}, J.},
   title = "{Can the Perturbation of a Stellar Motion in a Triple System Mimic
             a Planet}",
-   crossref={TDUG},
+  booktitle = {ESA SP-576: The Three-Dimensional Universe with Gaia},
+  year = 2005,
+  editor = {{Turon}, C. and {O'Flaherty}, K.~S. and {Perryman}, M.~A.~C.
+            },
+  month = jan,
```
timj: Latex was having a lot of trouble resolving the crossref to the other bib file so rather than debugging this I explicitly added the book details.

texmf/bibtex/bib/refs_ads.bib
```diff
@@ -1247,18 +1247,19 @@ @ARTICLE{2006MNRAS.369..677P
...
    eprint = {astro-ph/0302460},
+ keywords = {cosmology: dark matter, Galaxy: disk, Galaxy: halo, ISM: clouds, ISM: molecules},
      year = 2003,
     month = dec,
    volume = 412,
     pages = {105-120},
       doi = {10.1051/0004-6361:20031478},
-   adsurl = {http://adsabs.harvard.edu/cgi-bin/nph-bib_query?bibcode=2003A%26A...412..105M&db_key=AST},
-  adsnote = {Provided by the Smithsonian/NASA Astrophysics Data System}
+   adsurl = {http://adsabs.harvard.edu/abs/2003A%26A...412..105M},
```
timj: I re-downloaded this entry from ADS as it was causing strange problems. The key did change from `%26` to `&` so I'm not sure if that is a problem. I note that the ADS URLs are much simpler now so we should consider changing them over (it would be very easy to dynamically generate this file by sending a list of bibcodes to ADS and retrieving updated entries).

texmf/bibtex/bib/refs_ads.bib
```diff
@@ -1766,19 +1767,6 @@ @BOOK{2001sccd.book.....J
   adsnote = {Provided by the SAO/NASA Astrophysics Data System}
 }
 
-@ARTICLE{2003.Holland.IEEE,
```
timj: This entry was also in `refs.bib`.

PR #7, opened 2017-03-22
 added RHL beamers

womullan:
should not impact anything else but I will start to use these beamer classes instead of the one I used previously.  ..


timj:
These changes break Travis as the examples need `xparse.sty`. I think I need to add this package to the travis install section. Also, I tried building the new examples on my machine and they fail. I think examples should build otherwise they are not good examples. One question I have: if `LSSTcover` exists in multiple locations in `TEXMFLOCAL` how does it decide which to load?

RobertLuptonTheGood:
0. I sent these to Wil for a non-production version (I thought), so wasn't expecting testing yet.
1. Sorry,  xparse must be on my machine and also on Wil's
2. They build for Wil and me (using xelatex for the ones that manipulate fonts -- we'll have to deal with this when you sort out how to do the sans serif changes)
3. They look for LSSTcover and LSSTmain in LSST-themes/XXX as specified.  Is the README unclear?

timj:
Everything is tested 😄 

I can add `xparse.sty` so that's not a problem (but explains the failures above).

My breakage is this (with `xelatex`):

```
! LaTeX Error: File `LSST-themes/DESC/LSSTcover' not found.

See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.13 \maketitle
```
Hence my question. Looking deeper, it seems that the `ls-R` file lists these but the `DESC` directory has been missed from the PR.

I really do think that examples should work so that people can tweak them and know that breakage is something they did. I'm not sure which bit of the README explains this (I don't see anything in the PR). The README in the `examples` directory should probably be edited to explain what these examples are for.

I will need to tweak the Makefile if `xelatex` is required.

RobertLuptonTheGood:
Well, they all passed when I committed the code to git@github.com:mjuric/lsst_texdoc

timj:
https://github.com/mjuric/lsst_texdoc/tree/master/texmf/tex/LSST-themes/DESC exists there but not in this PR.

womullan:
mktexlsr should probably be part of the build - I did not checkin the
ls-r its not part of the repo .. I can add it now
but I think it needs to be constructed on the travis machine

Robert I put them in a branch and thought I got it all correct ...
indeeed worked on my machine !

timj:
@womullan I think that means that you didn't `git add` all the files in your local tree.

timj:
`ls-R` is in the repo history and this PR had changes to it.

womullan:
that is in the readme-lsst-beamer which I forgot to add done now

womullan:
mm  git status says only ls-R not added ...

timj:
I reverted your `ls-R` commit as you had put it at the wrong position in the directory structure. I'll bring the missing files over from the other repo and get them building with xelatex on travis.

timj:
Quick comments:

* Does this PR mean that `lsstbeamer.cls` should be removed as we now have the new way described here?
* I think the problem with the missing files is that the `.gitignore` ignores all PDF files so it didn't notice them sitting there in the DESC directory.
* Why are we shipping DESC styles?

timj:
Follow up, `xparse.sty` does not seem to be part of Ubuntu 12.04 (it is part of 14.04) and that's a problem because we are running 12.04 on travis. `texlive-xetex` does seem to exist for 12.04 but we'll have to disable Calibri font.

RobertLuptonTheGood:
It's a bit rude, but we could include xparse.sty in this package.  Or I could fix the code to use something else (might be painful -- raw TeX logicals are pretty bizarre, as I suspect, is Knuth)

womullan:
I have done that before.. just put it in the texmf tree.On 23 Mar 2017 1:56 a.m., Robert Lupton <notifications@github.com> wrote:It's a bit rude, but we could include xparse.sty in this package.  Or I could fix the code to use something else (might be painful -- raw TeX logicals are pretty bizarre, as I suspect, is Knuth)

womullan:
I would remove the lsstbeamer.cls when i update my existing presentations.
Perhpas a ticket on me to do this by May 1?

I do not see a problem having the DESC style ..

timj:
I updated Travis distribution. Now it's failing at the `ulem` package (and I also presumably need to remove the Microsoft fonts). I'll keep chipping away at this one.

timj:
The situation now is that everything is mostly working, except that if we want `xelatex` font support it seems like we have to replace Calbri (and Gill Sans) with OpenSans, and Palatino with TexGyrePagella. These fonts do seem to be in standard latex distributions but it looks like `xelatex` ignores TeX fonts and looks for system fonts. I think this means we have to put the two fonts in our `texmf` directory.

RobertLuptonTheGood:
I'm fine with dumping Calibri if it looks good-enough.

examples/README-LSST-beamer.md
```diff
@@ -0,0 +1,61 @@
...
+
+To get started put the directory containing `LSST-beamer.sty` and `LSST-themes` on your search path (e.g. add
+it to ``$TEXINPUTS`). These are in the `texmf` folder so if you have run `mktelsr` (or it is up to date in the repository) they will already be found.
+
+
+You can use beamer to make LSST-flavoured talks by adding something like:
+```latex
+\usepackage[fonts=false]{LSST-beamer}
+```
+to your beamer presentation (see `examples/example.tex` for a complete, umhh, example).  You need the
```
timj: We don't seem to have `example.tex` here and it's not in the `lsst_texdoc` repo either.

RobertLuptonTheGood: It should be in templates, probably as templates/beamer.tex.   Sorry about that.

timj: If it was `present_template.tex` then that is now `Example_presentation.tex` so I'll change it to that.

timj:
I merged this to get it out the way. We can fix up the DESC formatting issues in a later PR.

Issue #6, opened 2017-03-21
Sans serif fonts


mjuric:
A question about Helvetica (is it?) in our LaTeX — there’s something with the font that makes it look unpleasant.

I can’t really put my finger on it, whether it’s the kerning, the line spacing, or the rendering, or something else, it just looks odd. I think line spacing is a part of it — e.g., the author list on the front page of https://lse-163.lsst.io/v/dmsr-refresh/dpdd.pdf seems all squished?

Question: can we teach our LaTeX style to use Cambria (\usepackage{fontspec} \setmainfont{Cambria} \setmathfont{Cambria Math})? I think that would make these docs nearly identical to their Word counterparts.

(n.b. I personally don't mind the good old 'Computer Modern'...)

timj:
I don't really care all that much about the font but the project clearly prefers sans serif so I picked a sans serif font. I'm happy to switch to any other sans serif font. I'll try your suggestion.

timj:
`fontspec` seems to require `xelatex`. Are we okay with requiring that?

timj:
How are we expecting people to obtain `Cambria` and `Cambria Math`? They also have serifs so if we are going to require a Microsoft font we should use the official project font of `Calibri` (but there is no math font). These are all Microsoft fonts and require Office to be installed. This is going to cause problems in general (you can find the fonts online if you look hard enough and declare that you have bought Office at some point) and makes it impossible to build our documents on Travis.

timj:
Regarding `xelatex` the main problem I've run into (apart from removing packages we no longer need) is that it seems to define `\R` and `\B` and these clash with the versions we use in LSE-163 to indicate new and old text. I would have to use new names in LSE-163 but that's not a deal breaker.

timj:
How about OpenSans (which I seem to be able to get without using `xelatex`)?

RobertLuptonTheGood:
You can use {{\renewcommand}} to override xelatex's commands if we really don't need them (and use {{\let}} to rename the old ones)

mjuric:
Thanks Tim (and sorry to make you work so much -- it's a "nice to have" in the grand scheme of things; I did not mean to turn it into a research project!).

That said:
* OpenSans appears nice (based on 10 seconds of googling). Would it be hard to compile DPDD with it, to see what it looks like?
[...]

PR #8, opened 2017-03-24
DM-9930: Update font handling of lsstdoc

timj:


timj:
I have sent an updated DPDD example to @mjuric based on the discussion in #6 and he has indicated that he is happy with the improvement from this PR. I will treat that as permission to merge.

PR #9, opened 2017-03-27
DM-9941: Add CCB information to template

timj:
Add the CCB information to the title page and footer. CCB mode is derived by looking at the document handle (LDM, LSE etc). CCB information does not appear on title page in draft mode, but does appear in footer.

Required some adjustments to page layout to accommodate the new footer design.


timj:
This change raises the question of what to do with the Gaia "legacy" information (revision, status, etc) that no longer fit on the title page and end up on the second page with this change. Historically, for LSST the revision goes in the change record and the status is of draft/not draft is governed by the CCB text on the title page. Draft is also pretty obvious with the text in the background. If we have DM Science Notes in latex then we might end up with a different solution but I'm edging towards getting rid of the legacy items.

PR #11, opened 2017-03-31
Add standard baseline LSST documents to bib file

timj:
* Includes all the documents explicitly listed in lsstdoc.cls
* Adds \dscite command, like \ds but adding a reference entry.


timj:
Now that I'm using this in LDM-151, I wonder whether `\SRD` should result in `SRD, LPM-17` rather than `SRD`. The problem being that clicking on `SRD` takes you to LPM-17 rather than to the references entry (which is what happens for other references).

PR #12, opened 2017-04-03
DM-10072: Arabic page numbering in TOC-less document

ctslater:


timj:
The original code was meant to be how it worked. If you aren't using the `toc` option then you are responsible for closing out the front matter because you might be putting the table of contents in yourself a bit later. If we change expectation such that you can't put your own front matter in after `\maketitle` then this change can stand.

timj:
I should add that `toc` was not allowed to be the default for backwards-compatibility reasons (I couldn't break older documents). I'd be more than happy to make `toc` the default with `\setChangeRecord` mandated and then sort out the page numbering properly and not allow anything else in the preamble.

ctslater:
Ok, I can see how it makes sense to give people a way to add more to the front matter, but it's a little confusing that the TOC is the switch that does that. Converting back to Arabic manually would not have crossed my mind. I really don't have a strong opinion on a one-liner though, so whatever you want to do :).

timj:
I'm a bit stymied by backwards compatibility at this point. We could do one more round of sanity clean up if we committed to fixing up all the existing documents (~ 10 I would guess).

timj:
@ctslater do you want to propose a clean up option? We are running out of time as more people adopt the latex class.

timj:
@jdswinbank discovered this same problem with DMTN-044 today. We have to decide whether we are prepared to break all the existing documents to fix this? The obvious answer is to make `toc` the default state (who wouldn't want the table of contents to appear by default and the page numbering to reset?) and add a `notoc` option that just removes the table of contents and a new `\setDocPreamble` command that will let you specify arbitrary text that will be inserted during the roman numbering section.

@womullan I think we could do something like this if we adopt @jdswinbank's proposal that we recommend documents are built with submodules containing the texmf version that is known to work. I'll file an issue on that for comment.

timj:
I'm going to close this. I think we've run out of time to resolve the `toc` issue.

PR #15, opened 2017-04-07
DM-10139: Switch to an LSST bib style

timj:
* Default to `lsst_aa`.
* Use `@DocuShare` entries for LSST DocuShare. Stop using LiveLink.
* Support LiveLink in `lsst_aa`.
* Remove whitespace from DocuShare reference output.

Also, I'm not sure why we really need the URL field to be filled in. Can't the bst file work that out itself based on the number?


timj:
These changes will break LSST documents that use the shared `lsst.bib` and `gaia_aa` until the `gaia_aa` line is removed from the latex file (because it won't know about `@DocuShare`). The fix is straightforward for LDM-151 and LSE-163 as I have active PRs for those, but one option to mitigate this is to remove `gaia_aa.bst` and replace it with a soft link to `lsst_aa.bst`. @womullan would you prefer that in the short term?

timj:
I "backported" the docushare fixes to `gaia_aa` so that it won't break existing documents.

timj:
Turns out that documents explicitly setting `\bibliographystyle` now break because I, not unreasonably, force the style in `lsstdoc.cls`. Shall I stop doing that and force each document to see the bib style directly or live with the short term breakage?

jdswinbank:
So I know I asked this before, but I can't quite remember what the resolution is — can we just use submodules for this? It seems like you're jumping through hoops to make all possible documents work, whereas in fact just having each document specify a version of lsst-texmf it was happy with would avoid the problems.

Other than that, I've no objection to what's going on here.

timj:
Ok. I'm happy to give submodules within submodules a try (and it only really affects the people that want to do that for their particular repos, and people using `dm-docs`). It seems that `git clone --recursive` and `git submodule update --recursive` are the magic commands that will go all the way down (allowing `dm-docs` repo to work -- but I think I'll have to provide a couple of scripts in `dm-docs` for helping people remember this).

timj:
I talked to @womullan about the "breakage" problem and he's okay with taking the hit of having to remove the one line in existing documents. We are hoping that the class will settle down a bit in the near future so submodules won't be necessary. Remind me I said this in the summer when we are still breaking documents. @womullan has indicated that I should clean up the Bibtex Docushare item to remove the short and long number distinction (Since LSST handles are short). I'll fix that after I get #17 in a reasonable state.

Issue #16, opened 2017-04-07
adding a compiled example of LDM-nnn.pdf


suberlak:
Hi! 

I tried to add a compiled example of LDM-nnn.pdf   to the examples folder ,  but I have no write access.  It would really help to rename it as  Example_doc_LDM.tex,  and Example_doc_LDM.pdf  . The other files are labeled as examples and I looked for a while for an example of a document rather than a presentation.  It also took  4 people in my department to figure out that the texmf directory from your repo has to be put  to  ~/Library/texmf,   making sure that kpsewhich -var-value TEXMFHOME   points to ~/Library/texmf.    It is highly nontrivial   if one has never done anything like that ,  and I suggest putting this information on the front page .  

Best wishes
Chris

timj:
I've added "Data Management" team to this repo so you should be able to make a pull request now.

I specifically don't want compiled PDFs in the repo as it will balloon the size of the repository given an entirely new commit is required each time. I'd like the PDFs that get built by Travis to be persisted somewhere and I should talk to @jonathansick about that.

I  will think about renaming the example. The `Example_...` naming convention came after I had made the LDM example. I thought that `examples/LDM...` made it clear it was an example LDM but maybe that's because I already know what an LDM is. I don't really think that `example/Example...` adds any clarity but I'll take another look.

Also, I'm not sure I understand your compilation issue. Setting `$TEXMFHOME` to `~/wherever/lsst-texmf/texmf` is all that is needed to get things to work for me. It doesn't need to be installed into `~/Library`. I'll update the README to suggest `$TEXMFHOME` rather than `$TEXMFLOCAL` but they are more or less the same if neither have been set previously.

jonathansick:
I like the idea of having Travis build the examples and publish example products + documentation (essentially the README) on lsst.io — `lsst-texmf.lsst.io` maybe?

Not only will that keep pdfs out of the repo, it'll also add continuous integration that ensures the examples build.

timj:
They already build, but they aren't persisted. Master is protected and PRs are only allowed if Travis passes.

suberlak:
Hi!

[...]

PR #17, opened 2017-04-10
DM-10144: Add latex user guide

timj:
Adds a user guide describing how to use the latex classes and how to update them. The beamer documentation will be moved out of the README.

Note this uses rST to match the project style for software guides. It currently uses the DMTN stub code and needs to have the Sphinx boiler plate changed to use the standard scheme.

The documentation will refer to the example files. When the documentation is published, the example file PDFs will be included with the built documentation tree. @jonathansick is thinking about this.


timj:
@jonathansick here's a first stab at some latex style documentation. It includes special code for supporting DocuShare bibtex entries in `conf.py` that should probably be moved. Ideally `pybtex` would read the bst file directly but I'm not sure how to do that from sphinx.

Travis does build everything but nothing gets installed. Also not sure how to reference the example files from rST.

I'll leave this for now and we can talk about the strategy for publishing these docs. I'm happy to merge what I've got to `master` if people would rather get docs in the repository sooner rather than later before the web site is up and running.

jonathansick:
@timj Thanks. The new stuff by `pybtex` looks good and I'll get that into `documenteer`.

I think you can go ahead and merge this now and I'll pick it up and do the deployment work on a separate ticket.

timj:
@jonathansick I've merged this so that people can find some documentation (although I realize I forgot to mention the docs in the README. I've closed #16 since this includes some notes on the examples.

PR #21, opened 2017-04-12
DM-10144: Add \citeds command

timj:
`\citeds` should now be used for docushare and replaces `\citell` and the internal `\dscite` commands. `\citell` is now deprecated.


womullan:
sorry for not adding a new alias .. of course cited is fine and a citedp perhaps.

timj:
I added `\citeds` and `\citedsp` and support `\citeds[SRD]{LPM-17}`.

PR #23, opened 2017-04-15
Update README.md

mjuric:
Make the instructions on setting $TEXMFLOCAL more prominent.

The present README stumped ~four users (and led them astray to discover commands in TeX they never knew about :)).


timj:
Thanks. I'll tweak this to use `TEXMFHOME` as is written in the (new) user guide, then I'll merge.

PR #24, opened 2017-04-16
Bib entries for LDM-503

womullan:
just some typos and a ref for LDM-294

### DIFF
--- a/texmf/bibtex/bib/lsst.bib
+++ b/texmf/bibtex/bib/lsst.bib
@@ -18,7 +18,7 @@
   author =       {George Angeli and Robert McKercher},
   title =        {Change Controlled Document Cover Page and Style Guide},
   year =         2013,
-  month =        oct,
+  month =        Oct,
   handle =       {Document-9224},
 }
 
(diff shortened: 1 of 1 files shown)


texmf/bibtex/bib/lsst.bib
```diff
@@ -10,23 +10,23 @@ @DocuShare{Document-9224
   author =       {George Angeli and Robert McKercher},
   title =        {Change Controlled Document Cover Page and Style Guide},
   year =         2013,
-  month =        oct,
+  month =        Oct,
```
timj: I'm pretty sure this change isn't really needed because `oct` is defined as a macro at the top of the `lsst_aa.bst` file.

timj:
Quick comment that the usual branch naming convention is `u/user/something` or `tickets/DM-nnnn`. In this case we would expect `u/womullan/bibforldm-503` as the branch name.

womullan:
got it.

timj:
I've rebased this branch and made the use of month macros consistent. It was falling behind a bit.

timj:
I've merged these cleanups to make it easier to build LDM-502. We can always add more later.

PR #25, opened 2017-04-17
Support authoryear citations

timj:
Add `authoryear` class option.


timj:
@suberlak does this help you out? I added an example file and everything seems to work as I would expect.

timj:
This is working reasonably well for me so I've merged it. Hope it helps.

Issue #27, opened 2017-04-19
Remove non-standard document status commands


timj:
The style we inherited from Gaia has some features that are not used at LSST. These features are enabled with:

* `\setDocIssue`
* `\setDocStatus`
* `\setDocRevision`
* `\setDocAffil`
* `\setDocCurator`
* `\setDocApprove`
* `\setDocAuthorize`

[LPM-51](https://ls.st/LPM-51) indicates that Issue, Revision and Approve should be part of the Change Record, and that Status is either approved or draft (covered in the class by the `lsstdraft` option). Authorize seems to be handled by the change control process and is implied by knowing the document series. Curator has no obvious meaning at LSST.

@womullan Can we remove these (leaving the public API as a stub for now, possibly a latex warning) and clean up the class to remove them from `\maketitle`? Status and Revision are currently mandatory and lead to every latex document looking non-standard.

womullan:
ok the issue and revision appear on every page so if you do not set them how do I konw the issue and revision of this document ?  Having it in the change record is not sufficient - I wanted (had) that on every page though I see you have already removed that from the page header now. Perhaps the date as it appears now is better .. 

Affiliation is anyway optional and could be removed. 

Curator I would say remains useful .. it is optional  for large documents where the author list is long and we want to know  "who owns it?" the curator ...

Indeed Autorize ad Approve and status are not needed .. did you want me to modify the cls file ?

timj:
LSST style does not have issue and revision on each page, only on the change record. The question is whether we are going to insist on being different with latex or are matching the Word template and LPM-51. Issue is not a LSST concept. Release seems to be the specific revision. Data in the header is the project standard.

Curator may well be useful but we need to work out where to put it such that it doesn't look out of place. How about it goes at the end of the change record? (rather than finding somewhere for the title page)

I can remove the cruft.

womullan:
Ok we go LPM-51 - dates are usually good enough possible better...

Curator in the change record is fine - that also removes the need for it
as it was only set to PUT it on the front page.

timj:
This was fixed in #29.

PR #29, opened 2017-04-20
Remove Gaia document commands

timj:
LSST style does not require Authorize, Revision, Issue, Status, Affil or Approve. These have been removed and replaced by class warnings.

Authorize is implied by the document series. Approve, Revision and Issue are handled by the change record. Affil is not used. Status is released by default, or draft if the lsstdraft option is enabled.


timj:
This implements #27.

PR #30, opened 2017-04-21
DM-10311: Fix page layout and header spacing anomalies

timj:
This commit splits the header definition into two parts. The left header is the logo and horizontal rule and the right header is now the document description text. Previously they were a single left header but this led to spacing anomalies between pages where the spacing between the top of the logo and the top of the page could vary and the spacing between the ruler and the document identifier. Separating these into two unrelated entities seems to have stabilized the spacing issues.

In order to simplify the investigation, the page layout has now been defined using the geometry package rather than distinct settings that may interact unexpectedly. This has the side effect that the margins can now easily be displayed for debugging purposes using the `showframe` option.

All page layout was tweaked to make the spacing correct for the header. This led to the footer spacing now being more consitent, although page numbers do seem to move around from page to page.

### DIFF
--- a/texmf/tex/latex/lsst/lsstdoc.cls
+++ b/texmf/tex/latex/lsst/lsstdoc.cls
@@ -361,7 +361,7 @@
     }%
 
    \setlength{\parindent}{0pt}
-   \vspace*{-5cm}
+   \vspace*{-4cm}
    \putlogo
    \vspace*{1cm}
    \begin{center}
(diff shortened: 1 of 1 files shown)


texmf/tex/latex/lsst/lsstdoc.cls
```diff
@@ -581,46 +582,42 @@
...
-\setlength{\topmargin}{-0.5cm}
+\pagestyle{fancy}
 
+% Clear all headings
+\fancyhead{}
 \fancyheadoffset{0pt}
-\lhead{
-\color{lsstblue}
- \vspace{-4cm}
-  \IfFileExists{lsst_logo_notext.png}{
```
timj: I have removed this check on the basis that the image is always going to be there. I can add it back if required.

timj:
@jdswinbank thanks for the fixes. Inspired by your switch to a minipage for the left header I've now updated the header code to dynamically calculate where the text should go and how long the left ruler should be to align everything (the right ruler now simply fills the minipage that remains). Changing the logo width or changing the text below the logo will now work automatically.

PR #33, opened 2017-04-23
Properly set the header width.

jdswinbank:
This means we can run into the margins without generating warnings about
over-full hboxes.


timj:
(I'm trying to ignore the unholy mix of inches and centimeters (and points, and em) in this class.

PR #34, opened 2017-04-24
Adjust some overfull hboxes

timj:
In the preamble we currently have two overfull hboxes:

* The title bar under the logo.
* The change record table.

This pull request fixes the change record table and recalculates the title bar to dynamically fit the width of the paper and the width of the title text. As written there is still a complaint of 1inch of overfull hbox from the title (the left margin is 1 inch) and I'm not sure how to tell latex that the box with the logo and title bar is not subject to the normal page margins.


timj:
I rewrote the title page layout to directly place the logo and ruler using `eso-pic` just as we were already doing for the background image and thick ruler at the bottom of the page. This gives us direct control over placement and removes the underfull hbox warnings. The `\putlogo` command is removed (which I don't believe is part of the public API). I've also replaced the LSST logo with a non-blurry version.

PR #35, opened 2017-04-24
DM-10192: lsst-texmf.lsst.io website

jonathansick:
https://lsst-texmf.lsst.io/v/DM-10192


jonathansick:
> Would it be possible for the example (not tests) latex source to be visible inline? One of the important points of an example is that people can see the example and not just the rendered version.

I think so. I might turn examples.html into a landing page that links off to pages for each example. I think it might be too much to have the source for all examples on one page.

> Also, if I click on the PDF links for the examples it downloads them rather than showing them in browser (which is what a normal PDF link does for me).

I think this is due to the `:download:` role I was using there.

Will do.

jonathansick:
@timj Take a look at https://lsst-texmf.lsst.io/v/DM-10192/examples/index.html and pages therein.

PR #40, opened 2017-05-02
Add more refs required by LDM-294.

jdswinbank:


texmf/bibtex/bib/lsst.bib
```diff
@@ -277,6 +285,14 @@ @DocuShare{LDM-492
   handle =       {LDM-492},
 }
 
+@DocuShare{LDM-493,
```
timj: This document doesn't technically exist other than a placeholder on DocuShare.

jdswinbank: I understood this, assuming I could as "all".

LDM-294 contains some "forward references" to documents which don't really exist yet.

PR #43, opened 2017-05-05
Add EUPS table file to allow setup in-place

timj:


ups/lsst_texmf.table
```diff
@@ -0,0 +1 @@
+envPrepend(TEXMFLOCAL, ${PRODUCT_DIR}/texmf)
```
timj: @TallJimbo `TEXMFHOME` rather than `TEXMFLOCAL`?

TallJimbo: If that's what you'd advise, it's fine with me.  I've been using `TEXMFLOCAL` successfully but I don't know the distinction.

timj: See https://lsst-texmf.lsst.io/install.html for the current standard. `TEXMFLOCAL` is meant to point to a shared user directory tree of latex files. `TEXMFHOME` is a per-user collection of latex files. They have the same effect. Now that I explain that, maybe in EUPS land `TEXMFLOCAL` is the correct version to use if you think of EUPS installations as shared infrastructure... @jdswinbank might have an opinion.

timj: This link has details https://www.tug.org/texlive/doc/texlive-en/texlive-en.html#x1-110002.3

TallJimbo: What does the definition of `TEXMFHOME` mean when it says "The expansion of this variable dynamically adjusts for each user to their own individual directory"?  That sounds like it's not supposed to be an absolute path, and I think `${PRODUCT_DIR}` is going to be absolute.

timj: It means the default adjusts relative to `$HOME` because it's a per-user environment variable. It's an absolute value (or relative but that's generally confusion if you move around in your tree).

PR #47, opened 2017-05-09
Always get latest setuptools in Travis

jonathansick:
Normally Travis provides setuptools with its Python environment. In this case it seems we were getting an old, and possibly broken, version of setuptools. In the Travis install phase we force an upgrade of
setuptools; this solves the build issue:

```
making output directory...
Exception occurred:
File "<environment marker>", line 1, in <module>
  NameError: name 'platform_system' is not defined
```

Also passed the -T flag to Travis to get more verbose exception tracebacks.


jonathansick:
I think Travis's trusty VM is considered in "beta" so I wonder if they made some inadvertent change to the image.

PR #55, opened 2017-06-01
DM-10641: Add cookiecutter templates for documents and presentations.

jonathansick:
Adds two [cookiecutter](https://cookiecutter.readthedocs.io/en/latest/index.html) project templates. One for a lsstdoc document, another for an LSST-beamer presentation.

Docs: https://lsst-texmf.lsst.io/v/DM-10641/templates (I also improved other reference pages to cross-link to the template pages).

Note: we'll need to revisit the document template documentation (https://lsst-texmf.lsst.io/v/DM-10641/templates) to better link to process documentation. The lsst-texmf docs should be responsible for documenting the technical usage of the template but not the organizational policies. That page will also need to be changed with the `project create` chatbot service comes online.

### DIFF
--- a/docs/beamer.rst
+++ b/docs/beamer.rst
@@ -4,7 +4,12 @@
 Writing Presentations
 #####################
 
-You can use `beamer <https://en.wikipedia.org/wiki/Beamer_(LaTeX)>`_ to make LSST-flavoured talks by adding something like:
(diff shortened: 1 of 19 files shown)


jonathansick:
Clearly, I go the extra mile for my LaTeX user base.

PR #75, opened 2017-07-05
Add docs on use as a submodule.

ktlim:


ktlim:
@jonathansick This may also be appropriate for the cookiecutter README.rst.

jonathansick:
We've decided to go with lsst-texmf as a submodule officially?

I'm not 100% sure of how to get cookiecutter to handle submodule setup, but it might be possible to have a [post-generation hook](https://cookiecutter.readthedocs.io/en/latest/advanced/hooks.html) do this. I'll also have to change the  [Makefile template](https://github.com/lsst/lsst-texmf/blob/master/templates/presentation/%7B%7Bcookiecutter.slug%7D%7D/Makefile) to use lsst-texmf from the submodule instead of from the Docker container's environment.

👉 https://jira.lsstcorp.org/browse/DM-11194

timj:
I don't think this PR says it's policy, it says that it's a convenient thing to do. Once the DM review documentation crunch is out the way I'm going to talk to people who have experienced the crunch to work out if we can agree on a doc development policy. That policy would go in the dev guide (I already have a draft there but it might be changed). One problem with the `lsst-texmf` as submodule approach is that it makes it confusing if you get travis to use the docker container but then aren't using the texmf from the docker container.

jonathansick:
> One problem with the lsst-texmf as submodule approach is that it makes it confusing if you get travis to use the docker container but then aren't using the texmf from the docker container.

I think this would be okay. If the `Makefile` sets up `TEXMFHOME` to point to the submodule, that should be clear enough.

PR #82, opened 2017-07-11
Change DM TCT to DMCCB.

ktlim:


ktlim:
You appear to check that the build succeeds without any check on content.  If LaTeX returned zero exit status but produced empty output, Travis would still pass, correct?

timj:
Thanks for this. Now that LDM-294 was baselined.

Regarding tests: yes, the test is "does latexmk run okay", not "is a usable PDF generated". You can check the lsst-texmf web site where the files are deployed.

PR #114, opened 2018-02-17
First attempt at python acronyms extraction

timj:
Not ready to merge yet but basic functionality exists.

### DIFF
--- a/bin/generateAcronyms.py
+++ b/bin/generateAcronyms.py
@@ -11,7 +11,12 @@ import os.path
 import sys
 import re
 
-MATCH_ACRONYM = "^(\w+)\s*:\s*(.*)$"
+try:
+    import pypandoc
+except ImportError:
+    pypandoc = None
+
+MATCH_ACRONYM = "^([\w/&\-\+]+)\s*:\s*(.*)$"
 MATCH_ACRONYM_RE = re.compile(MATCH_ACRONYM)
 
 
@@ -115,6 +120,7 @@ def read_myacronyms(filename="myacronyms.txt", allow_duplicates=False, defaults=
     with open(filename, "r") as fd:
         for line in fd:
             acr, defn = _parse_line(line)
+            print(line, defn)
             if acr is None:
                 continue
 
@@ -173,7 +179,7 @@ def read_skip_acronyms(file_name="skipacronyms.txt"):
     return skip
 
 
-def find_matches(filename, acronyms, ignore_str=" %"):
+def find_matches_per_line(filename, acronyms, ignore_str=" %"):
     """Return list of matching acronyms in file.
 
     Parameters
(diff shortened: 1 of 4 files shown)


womullan:
Just trying it .. lots of difficult to understand errors about files  and then it complains "RuntimeError: Duplicate definitions of CAM differ" It produces no new file .. I think it a=ought to produce the acronyms anyway better than nothing ..   even when I delete CAM from myacronyms.txt it still gives this error ..

womullan:
SO it seem to still look in myacronyms.tex as well as myacronyms.txt .. hence my problem above. Now it runs through . The Gaia version warns when I have multiple definitions this does not .. so I have in acronyms.tex  .. 
BW & BandWidth \\\hline
BW & Baade's Window (special, high-density area on the sky) \\\hline

this may be ok but the user should be warned to pick one to put in the myacronyms.txt .. I think also in Gaia we only put the FIRST one ..  so we have not so many duplicates (again warning the user about that).

timj:
Yes. I'll fix those. It's not ready for release yet but it does all the basics. The CAM problem is that there are two definitions in the LSST file in lsst-texmf/bin -- I could treat them like the Gaia ones and only complain if there is no clarification in the local file.

jdswinbank:
Do we really want something home-grown for this? Seems like the [glossaries](https://ctan.org/pkg/glossaries?lang=en) package would be adequate...

timj:
I did think of glossaries but @womullan explained why it didn't work for them in Gaia. I think part of the problem is that you have to declare explicitly that you are using a glossary term when you write the latex whereas the gaia approach is to scan the text trying to find acronyms. I do remember trying to upgrade LPM-17 to use glossaries instead of `glossary` and I gave up because it wasn't able to do what SRD wanted. I'm willing to be convinced.

jdswinbank:
I don't really want to convince people one way or the other — I used glossaries in DMTN-020, and they worked fine, and I'm cautious that rolling our own involves implementation effort up front (which it looks like you've already broken the back of) and ongoing maintenance/documentation/compatibility/etc issues, which are not free. If @womullan reckons this is a good trade off, I'm not actually going to argue (for a change).

womullan:
Tim has pretty much done the Gaia thing (still needs tweaking) - it did not require much maintenance on Gaia later.  Yes the problem with glossary is you have to  remember to \gls{} entries  and people are lazy or will simply forget - so then you need a script to check up on that. The acronym approach basically does most of the job for you on any document at anytime and warns about the bits you need to take care of - it also requires no modification of the actual tex..

timj:
I think these two commits were on the wrong branch and I can remove them.

womullan:
well we could apply them to master .. actually your script could go to well if not already done ..

timj:
I think @jdswinbank added the references separately in #117

PR #129, opened 2018-06-25
DM-14878: Report acronyms that have no definitions

timj:
Update the acronym scanner so that we separate non-standard acronyms ("T&S", "FoV") from all upper case acronyms. Then report acronyms (2 or more upper case characters) that we lack definitions for.


womullan:
It does not look at the skipacronynms.txt .. so includes ones in that file. Other wise great

timj:
It's meant to look at skipacronyms.txt. https://github.com/lsst/lsst-texmf/blob/master/bin/generateAcronyms.py#L315-L331

I'll have a look at it to see if it's really working.

timj:
I've fixed the skip reporting.

PR #140, opened 2018-10-31
Improve the generated makefile and add acronyms etc. Also short form for no controled docs (no title page)

womullan:

PR #141, opened 2018-10-31
Refactor submodule docs from installation

jonathansick:
At some point some notes on setting up lsst-texmf as a Git submodule got added to the installation page, which is good content but not the right place or presentation in my opinion.

This change:

- Creates a new page all about submodule usage (set up and cloning)
- Clarifies that the install page is all about creating a centralized installation and links to the docker and submodule pages for alternative approaches.

Preview:

- [Installing lsst-texmf](
https://lsst-texmf.lsst.io/v/u-jonathansick-submodules-docs/install.html)
- [Using lsst-texmf as a Git submodule](
https://lsst-texmf.lsst.io/v/u-jonathansick-submodules-docs/submodule.html)

I could also add notes on adding `bin/` to `$PATH`, but I'm not sure that'll be tackled in #140.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -13,6 +13,7 @@ lsst-texmf is on GitHub at https://github.com/lsst/lsst-texmf.
    :maxdepth: 2
 
    install
+   submodule
    docker
    examples/index
    templates/index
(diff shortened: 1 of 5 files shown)


womullan:
looks good

PR #145, opened 2018-12-10
Added ops and ts .. plus ops doc tcodes

womullan:


jonathansick:
That's a lot of new document codes! I've rebased this onto the fixed master and merged.

PR #150, opened 2018-12-19
DM-16905: Copy full lsst-textmf project to Docker image

### LINKED ISSUE #147: Preserve lsst-texmf directory structure in Docker image
`generateAcronyms.py` [relies on the directory structure](https://github.com/lsst/lsst-texmf/blob/29683d708f75bdc068ac0d09951a08e27dfbbf8f/bin/generateAcronyms.py#L335-L338) of lsst-texmf itself, which is fine. When the Docker image is constructed through, the `lsst-texmf` repo isn't copied as is, so that the `etc/` directory isn't available to `generateAcronyms.py`. This can be fixed if the `lsst-texmf` repo is copied in full into the Docker image.

jonathansick:
Rather than than trying to copy over just the parts of lsst-texmf that we think are necessary, this commit copies over lsst-texmf in full into `/lsst-texmf/` of the image. This prevents issues like #147 where `generateAcronyms.py` looks for a file in the `etc` directory of `lsst-texmf`.

The `.dockerignore` file helps to avoid adding truly unnecessary files.

Fixes #147

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,20 @@
+# Repo files not needed in the image
+.github
+docs
+.gitignore
+.travis.yml
+Makefile
+
+# Files built by tests
+*.pdf
+*.aux
+*.bbl
+*.blg
+*.fdb_latexmk
+*.fls
+*.log
+*.out
+*.pdf
+*.toc
+*.nav
+*.snm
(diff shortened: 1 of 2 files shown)


jonathansick:
This shouldn't affect existing usage since the `$TEXMFPATH` environment variable is still set in the container.

PR #153, opened 2019-02-01
Check that pandoc is available.

jdswinbank:
Pypandoc happily imports if pandoc isn't installed, but then raises when you
try to use it.

### DIFF
--- a/bin/generateAcronyms.py
+++ b/bin/generateAcronyms.py
@@ -13,7 +13,10 @@ import re
 
 try:
     import pypandoc
-except ImportError:
+
+    # raises OSError if pypandoc is available but pandoc isn't.
+    pypandoc.get_pandoc_path()
+except (ImportError, OSError):
     pypandoc = None
 
 MATCH_ACRONYM = "^([\w/&\-\+]+)\s*:\s*(.*)$"


jdswinbank:
@timj — this enables me to at least build LDM-294 with `generateAcronyms.py`, which I think is a useful step forwards.

More generally, though, I'm a bit uncomfortable about having two code-paths through here. Do we need a pandoc and a non-pandoc version? If the non-pandoc version is adequate, can we just use that and delete the pandoc version for simplicity? If the non-pandoc version _isn't_ adequate, can we delete it so that it doesn't get used by mistake?

timj:
`pandoc` is more correct since it does a much better job of filtering out latex markup before the regex runs looking for acronyms. Without pandoc you get a few more false positives (e.g., with words that look like they are distinct because they have accents in them). I originally assumed that this would be run by the author who would then manually deal with the incorrect entries and commit the generated tex file. I wasn't expecting the whole process to be automated. Anyhow, pandoc is an incremental improvement over no-pandoc but you may not notice a difference for a given document since it depends on the document contents.

jdswinbank:
Thanks for the quick review Tim. I still wish that one code path or the other would go away, but I'm out of free energy to care about it for the moment...!

PR #159, opened 2019-02-22
 2019 year in cookiecutter and some fixed to the body.tex

womullan:
just minor stuff

### DIFF
--- a/etc/glossary.txt
+++ b/etc/glossary.txt
@@ -31,6 +31,7 @@ BAC        : Budget At Complete
 BCWP       : Budgeted Cost of Work Performed
 BCWS       : Budgeted Cost of Work Scheduled
 BNF        : Backus-Naur Form
+BOE        : Basis Of Estimate
 bps        : bit(s) per second
 Bps        : Bytes per second
 CAM        : Control (or Cost) Account Manager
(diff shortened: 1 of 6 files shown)


jdswinbank:
Yes... but not all. I think we should only be carrying acronyms which are very generic or which are LSST-specific. Let's not store things like BBC, TSESO or MREN — they're just likely to cause collisions.

womullan:
we have a lsst specific gloassry for the other stuff -- i reduced the myacronyms a lot though now and moved some to the gloassary.txt  or lsstacronyms.txt in etc

jdswinbank:
Right — LSST specific abbrevs should go in `lsstacronyms.txt` and really generic stuff in `glossary.txt`. But having ”BBC” recorded *anywhere* by default just means that somebody will refer to the British Broadcasting Corporation and we'll end up with a glossary entry claiming it's a German shipping company and confusion will reign.

I reckon we should only add such easily-confused abbreviations to specific documents on a case-by-case basis.

womullan:
pretty much what i did - BBC is still there .. but it could be removed altogether if that is what you mean

jonathansick:
Right, what I'm getting at is that if an acronym isn't fit for lsst-texmf because it's not universal enough, it's probably for universal enough for the template either (I.e.,`macronyms.tex` should start empty).

jdswinbank:
I would remove BBC and anything else that isn't either LSST-specific or completely unambiguous in basically all circumstances from this repository (including both the `etc` directory and `myacronyms.txt` — which I think is what we're talking about, as `myacronyms.tex` will be autogenerated and shouldn't be versioned at all).

womullan:
the myacronyms here is from the template and is copied when you use cookie cutter to make a new doc. I will remove more and leave one example in there ..

PR #164, opened 2019-04-29
Genrateacronyms no does gloassary - still need to add gloassary terms to the gloassry.txt file

womullan:
there is also some funny interaction with our CLS file and gloassaries which needs to be resolved ..  but the script works to make a glossary file and \gls most things


bin/generateAcronyms.py
```diff
@@ -19,9 +32,16 @@
 except (ImportError, OSError):
     pypandoc = None
 
-MATCH_ACRONYM = "^([\w/&\-\+]+)\s*:\s*(.*)$"
+#DO glossary or not  (-g)
+GLS=False
+#Go through files on second pass and \gls  or not (-u)
+UPD=False
+
+#Match for extracting acronyms from the glaossry myacronyns .txt files
```
timj: Please run this through flake8.

bin/generateAcronyms.py
```diff
@@ -246,6 +266,7 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
         # Use markdown rather than plain text because
         # for plain text \textbf{Int} is converted to "INT"
         # for emphasis.
+        # \gls{XXX} however is comletely removed
```
timj: "completely"

bin/generateAcronyms.py
```diff
@@ -288,16 +308,26 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
     regex = re.compile(pattern)
 
     matches = set(regex.findall(text))
-
+    #print(matches)
```
timj: Remove debug print.

bin/generateAcronyms.py
```diff
@@ -288,16 +308,26 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
     # Now look for all acronym-like strings in the text, defined as a
     # collection of 2 or more upper case characters with word boundaries
     # either side.
     regex = re.compile(r"\b[A-Z][A-Z]+\b")
     used = set(regex.findall(text))
 
+    # now Glossary entries Gls gls use group ( ) to catch what's between { }
+    regex = re.compile("ls{([\w ]+)}")
```
timj: Is this regex too greedy if you have a scenario of: "blah \gls{something here} and \gls{over here} blah" ?

bin/generateAcronyms.py
```diff
@@ -288,16 +308,26 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
+    gls = set(regex.findall(text))
+
+    used = used.union(gls)
+    #print (gls)
+    #print (acronyms)
     # For all acronyms that were used and have existing definitions, add
     # them to the current list of matches
-    matches.update(used & acronyms)
+    matches.update(used & acronyms )
```
timj: Remove the extra space here.

bin/generateAcronyms.py
```diff
@@ -288,16 +308,26 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
     # collection of 2 or more upper case characters with word boundaries
     # either side.
     regex = re.compile(r"\b[A-Z][A-Z]+\b")
     used = set(regex.findall(text))
 
+    # now Glossary entries Gls gls use group ( ) to catch what's between { }
+    regex = re.compile("ls{([\w ]+)}")
+    gls = set(regex.findall(text))
+
+    used = used.union(gls)
```
timj: I think `used.update(gls)` will be better.

bin/generateAcronyms.py
```diff
@@ -288,16 +308,26 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
+    #print (gls)
+    #print (acronyms)
```
timj: Remove debug prints.

bin/generateAcronyms.py
```diff
@@ -288,16 +308,26 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
     # For all acronyms that were used and have existing definitions, add
     # them to the current list of matches
-    matches.update(used & acronyms)
+    matches.update(used & acronyms )
+    # For all gls entries  that were used and have existing definitions, add
+    matches.update(gls & acronyms )
```
timj: Weren't the gls entries already combined with `used` above? Does this do anything?

bin/generateAcronyms.py
```diff
@@ -288,16 +308,26 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
+    # still a probles for COMPOUND ENTRIES like NASA ROSES  .. ROSES is tagged as missing
```
timj: "probles" -> "problem"

bin/generateAcronyms.py
```diff
@@ -309,6 +339,22 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
 find_matches = find_matches_combo
 
 
+def write_latex_glossary(acronyms, fd=sys.stdout):
+    """ Write a glossary file with newgloassaryitem per definiton.
```
timj: newgloassaryitem has an extra "a".

Convention is not to use a leading space after the `"""`.

bin/generateAcronyms.py
```diff
@@ -309,6 +339,22 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
+
+    Parameters
+    ----------
+    acronyms : `list`
+        List of 2-tuples with acronym and definition.
+    """
```
timj: Technically also need to document fd:

```
fd : `file`, optional
    Output stream to use.
```
I realize I forgot myself later on.

bin/generateAcronyms.py
```diff
@@ -309,6 +339,22 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
+
+    print ("% DO NOT EDIT - generated by "+sys.argv[0]+ " from https://lsst-texmf.lsst.io/.", file=fd)
```
timj: No space between print and the paren (flake8 will catch that).

bin/generateAcronyms.py
```diff
@@ -309,6 +339,22 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
+    for acr, defn in acronyms:
+	# why \ for some stuff but {{ for {
```
timj: `{{` is because `{` has to be escaped when using `.format()`.  `\\` is because you have to escape the backslash unless you use a `r""` raw string.

bin/generateAcronyms.py
```diff
@@ -309,6 +339,22 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
+        print("\\newglossaryentry {{{}}} {{name={{{}}}, description={{{}}}}}".format(acr, acr, defn), file=fd)
+
+       #\newglossaryentry {APL} {name={APL}, description={ Apache Public License }}
```
timj: If this is an example of what the output line should look like then it's probably better to put it before the print.

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
         else:
             raise RuntimeError("Internal error handling {}".format(acr))
 
-    with open("acronyms.tex", "w") as fd:
-        write_latex_table(results, fd=fd)
+    if (GLS):
```
timj: GLS should be an argument to `main()` and not a global variable. Similarly UPD.

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+       with open("aglossary.tex", "w") as gfd:
+           write_latex_glossary(results, fd=gfd)
+    else:
+       with open("acronyms.tex", "w") as fd:
+           write_latex_table(results, fd=fd)
+
+def loadGLSlist() :
+   fname="aglossary.tex"
+   fin = open (fname,'r')
```
timj: Use context manager so that the file gets closed in all cases. (see `with open` a couple of lines above).

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
-    with open("acronyms.tex", "w") as fd:
-        write_latex_table(results, fd=fd)
+    if (GLS):
+       with open("aglossary.tex", "w") as gfd:
+           write_latex_glossary(results, fd=gfd)
+    else:
+       with open("acronyms.tex", "w") as fd:
+           write_latex_table(results, fd=fd)
+
+def loadGLSlist() :
```
timj: Please add at least a one line docstring saying what this is for.

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+   fin = open (fname,'r')
+   # match gls entry but only take the first group
+   regex = re.compile(r'{(.+)}\s*{.*{.+}.*description={.+}}')
+   text = fin.read()
+   GLSlist=set(regex.findall(text))
+   return GLSlist
+
+
+def glsfn(s):
+   "put \gls{} -- used in the regexp substitution"
```
timj: Convention is to use `"""`

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+def updateFile(inFile,GLSlist):
+    """Update the tex file by looking for acronyms and glossary items GLSlist."""
+    nf=inFile
+    of = nf.replace(".tex",".tex.old")
+    sh.copyfile(nf,of)
+    fin = open (of,'r')
+    fout = open (nf,'w')
+
+    for line in fin :
+        if not line.startswith('%') : #its a comment ignore
```
timj: "it's" (also flake 8 violation but I won't comment on those since flake8 is better at it than me).

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+   "put \gls{} -- used in the regexp substitution"
+   #print (s)
+   return s.group(1)+"\\gls{"+s.group(2)+"}"+s.group(3)
+
+def updateFile(inFile,GLSlist):
+    """Update the tex file by looking for acronyms and glossary items GLSlist."""
+    nf=inFile
+    of = nf.replace(".tex",".tex.old")
+    sh.copyfile(nf,of)
+    fin = open (of,'r')
```
timj: Use `with open`

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+
+
+def glsfn(s):
+   "put \gls{} -- used in the regexp substitution"
+   #print (s)
+   return s.group(1)+"\\gls{"+s.group(2)+"}"+s.group(3)
+
+def updateFile(inFile,GLSlist):
+    """Update the tex file by looking for acronyms and glossary items GLSlist."""
+    nf=inFile
```
timj: The variable names are a bit obtuse.

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+    of = nf.replace(".tex",".tex.old")
+    sh.copyfile(nf,of)
```
timj: I think we just need `os.rename` here. We are copying back to the original file and printing each line anyhow. We might need to trap a problem and rename it back to the original if something goes wrong in the loop.

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+    fin = open (of,'r')
+    fout = open (nf,'w')
+
+    for line in fin :
+        if not line.startswith('%') : #its a comment ignore
+            for g in GLSlist :
+               regx =re.compile("([,\s(](?<!=\\gls))("+g+")([)\s,'.])")
```
timj: Do not compiler a regex in the loop. The speed advantage is to compile once and then reuse it.

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+               res =regx.search(line)
+               if (res is not None) : #ok now .. more checks
+                  if (re.search(r"gls{.*"+g+".*}",line)) : #already glsed or part of another gls item
```
timj: I don't think this needs to be a raw string because there are no escapes.

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+
+def update(texfiles):
+    """Update the passed tex files by looking for acronyms and glossary items loaded from aglossary.tex."""
+
+    if not texfiles:
+        raise RuntimeError("No files supplied.")
+
 
+    print ("Updating gls texfiles  the orignals file will be .old ")
+    print ("If glosarray items contian \gls refs you may need to run this again to catch the entries in aglossary.tex ")
```
timj: Spelling in both these lines.

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+    parser.add_argument('files', metavar='FN', nargs='+',
+                    help='FILE to process')
+    parser.add_argument('-g','--glossary',action='store_true',
+                        help=" Generate aglossary.tex a glossary file for acronyms and glossary entries.")
+    parser.add_argument('-u','--update',action='store_true',
+                        help="Update files to put \gls on acronyms and glossary entries .")
+
+    args = parser.parse_args()
+    GLS=args.glossary
+    UPD=args.update
```
timj: I don't think this is a global is it? Isn't it used a few lines below so could use `args.update` directly?

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+                  else : # .. find and add GLS - theoretically just the substitution - easier with a function
+                     line=regx.sub(glsfn,line)
+		     # print (str(res)+" "+g+" ->"+line)
+        fout.write(line);
+    fout.close()
+    fin.close()
+
+
+
+def update(texfiles):
```
timj: I think a clearer function name would be good. Something like `rewrite_files` ?

bin/generateAcronyms.py
```diff
@@ -410,11 +456,86 @@ def main(texfiles):
...
+
+    parser.add_argument('files', metavar='FN', nargs='+',
+                    help='FILE to process')
+    parser.add_argument('-g','--glossary',action='store_true',
+                        help=" Generate aglossary.tex a glossary file for acronyms and glossary entries.")
+    parser.add_argument('-u','--update',action='store_true',
+                        help="Update files to put \gls on acronyms and glossary entries .")
+
+    args = parser.parse_args()
+    GLS=args.glossary
```
timj: GLS should be passed into `main()`.

texmf/bibtex/bib/refs_ads.bib
```diff
@@ -5390,4 +5390,33 @@ @ARTICLE{2002cs........8011G
       adsnote = {Provided by the SAO/NASA Astrophysics Data System}
 }
 
+i
```
timj: This `i` seems wrong.

texmf/bibtex/bib/refs_ads.bib
```diff
@@ -5390,4 +5390,33 @@ @ARTICLE{2002cs........8011G
...
+@INPROCEEDINGS{2019AAS...23324505B,
```
timj: Are these really needed for this PR or were they part of something else?

bin/generateAcronyms.py
womullan: done

bin/generateAcronyms.py

bin/generateAcronyms.py

bin/generateAcronyms.py
womullan: I think not .. } is not a word   - it seems to work ..

bin/generateAcronyms.py
womullan: i guess i got it in the deflakeing

bin/generateAcronyms.py
womullan: yes it was going to be used elsewher but in the end I did not

bin/generateAcronyms.py
womullan: update(texfiles) looks pretty clear to me - you are being to nit picky

texmf/bibtex/bib/refs_ads.bib
womullan: oh yea but bibtex seems not tot care

texmf/bibtex/bib/refs_ads.bib
womullan: some thing else .. but let it slide ..

bin/generateAcronyms.py
```diff
@@ -19,21 +33,29 @@
...
+pypandoc = None  # it can not hangle gls
+
 
 def _parse_line(line):
     """Parse a line and return an acronym and definition.
 
     Parameters
     ----------
-    line : `str`
+    line: `str`
```
timj: numpydoc wants the space between parameter and the colon.

womullan: ok - inconsistent .. I begin to side with Robert on the pointlesness of some of this

bin/generateAcronyms.py
womullan: what is the point in making it global if i pass it in ...

womullan: removed

bin/generateAcronyms.py
womullan: no the gls are distinct from the acronyms this builds a complete list.

bin/generateAcronyms.py

bin/generateAcronyms.py

bin/generateAcronyms.py
womullan: added

bin/generateAcronyms.py
womullan: k

bin/generateAcronyms.py
womullan: yea should have done the excepiton handling .. added now

bin/generateAcronyms.py
womullan: ok - map of regex made before entering the loop

bin/generateAcronyms.py
womullan: does it matter ?

timj: No, but it might confuse a future reader of the code who is trying to guess why the string was special.

PR #166, opened 2019-05-15
single glossary with tags

womullan:


bin/generateAcronyms.py
```diff
@@ -75,6 +75,70 @@ def _parse_line(line):
...
+            hasTag = False
+            if tags:
+                tagset.update(tags.split())
+                hasTag = tagset.intersection(utags)
+            if acr is None:
+                continue
+
+            if acr not in definitions:
+                definitions[acr] = set()
+            # Ok lets try to do something with Tags .. like if its tagged take this one
```
timj: nitpick: There are lots of typos in these comments.

bin/generateAcronyms.py
```diff
@@ -75,6 +75,70 @@ def _parse_line(line):
...
+        name, definition, tags etc ..
+    tags : `set'
+        List of tags supplied by user to decide which definintion to keep when there are many.
+    init : `dict`
+        Initial definitions to augment with the content from this file.
+
+    Returns
+    -------
+
+    def acronyms : `dict`
```
timj: nitpick: remove "def"

bin/generateAcronyms.py
```diff
@@ -75,6 +75,70 @@ def _parse_line(line):
...
+    """
+
+    if init is None:
+        definitions = {}
+    else:
+        definitions = init.copy()
+
+    lc = 0
+    with open(filename, "r") as fd:
+        reader = csv.reader(fd, delimiter=',', quotechar='"')
```
timj: This is all fine but I wonder if we could tweak the CSV to make it readable by `astropy.table.Table.read()` with format "csv" so you would get a table with column names read directly from the file.

bin/generateAcronyms.py
```diff
@@ -561,13 +615,19 @@ def update(texfiles):
     parser.add_argument('-u', '--update', action='store_true',
                         help="""Update files to put \\gls on acronyms and
                                 glossary entries .""")
-
+    parser.add_argument('-t', '--tags',
+                        help="""Space seprated list of tags between quotes
```
timj: "seprated"

bin/generateAcronyms.py
```diff
@@ -561,13 +615,19 @@ def update(texfiles):
     parser.add_argument('-u', '--update', action='store_true',
                         help="""Update files to put \\gls on acronyms and
                                 glossary entries .""")
```
timj: nitpick: space before period.

bin/generateAcronyms.py
womullan: ok ran aspell through it .. i can not do more

bin/generateAcronyms.py

bin/generateAcronyms.py
womullan: perhaps for another day ..

bin/generateAcronyms.py
womullan: aspell got it

PR #169, opened 2019-06-03
script to scrape bib entry from file(s) in repo

womullan:


bin/generateBibentry.py
```diff
@@ -0,0 +1,137 @@
...
+    comment = re.compile(r"%.*")
+
+    # Read the content of the file into a single string
+    lines = []
+    year = ""
+    month = ""
+    auths = ""
+    titles = ""
+    handles = ""
+    meta = filename == "meta.tex"
```
timj: Not all our docs have a `meta.tex` file.

womullan: yes there is an if there ... if they do not have meta.tex I have no way to get the date from \tody for example.. should i put in the current month and year

womullan: Renamed script

PR #170, opened 2019-06-04
Update xits-math.otf filename for TeXLive 2019.

jdswinbank:
See e.g.
https://tex.stackexchange.com/questions/43642/xits-font-not-found#comment1142830_43680.

This (presumably) breaks for everybody who is *not* running TeXLive 2019, which is awkward. Should we actually be bundling the `.otf` inside lsst-texmf or something?


timj:
Are you saying that this works fine until we use 2019 and then it breaks because the math font changed name?

womullan:
Yes this works for john with texlive 2019 (and the current texmf does not work for him). Vice versa if I take this branch it does not work for me (older Tex live) ..

jdswinbank:
Ok, let's try adding the `XITSMath-Regular.otf` font. Cross fingers...

jdswinbank:
Looks like that makes Travis happy. Does it also please @timj & @womullan ?

PR #171, opened 2019-06-13
Add author handling tools for writing AAS commissioning papers

timj:
As we start to develop commissioning papers we need to develop the author management tools and put them in a standard location. In this pull request we copy the relevant files from lsst-pst/overview_paper#57


timj:
Recreated as #172

PR #172, opened 2019-06-13
DM-20072: Transfer author support tools from overview paper

timj:
In this pull request we copy the relevant files from lsst-pst/overview_paper#57

This is a resubmission of #171 with the correct ticketed branch name.


.travis.yml
```diff
@@ -8,6 +8,10 @@ matrix:
     - python: "3.5"
       # Enable docker and documentation deployments
       env: LTD_MASON_BUILD=true DEPLOY_DOCKER_IMAGE=true
+    - python: '3.6'
```
womullan: was it really necessary to force me to upgrade python .. ..:)

timj: This is the python to use to run flake8 inside Travis. DM only use 3.7 now so I haven't tried db2authors with an older python.

timj:
I won't merge until I get some feedback from @ivezic that he's okay with commissioning papers and the overview paper depending on lsst-texmf.

ivezic:
I am ok with your excellent plan!

PR #174, opened 2019-06-17
U/womullan/typosmissingglos

womullan:
do you want to review these type of chnages ?


etc/glossarydefs.csv
```diff
@@ -464,7 +466,7 @@ Subcontract,An agreement under which another entity will perform part or all of
 Subsystem,A set of elements comprising a system within the larger LSST system that is responsible for a key technical deliverable of the project.,Adm,,
 SSM,\gls{Subsystem Manager},Adm,,
 Subsystem Manager,"responsible manager for an LSST subsystem; he or she exercises authority, within prescribed limits and under scrutiny of the Project Manager, over the relevant subsystem's cost, schedule, and work plans",Adm,,
-Subsystem Scientist,The principal science advisor to a Subsystem Manager; he or she ensures that the subsystem specifications are appropriated for achieving the project's goals,Adm,,
+Subsystem Scientist,The principal science adviser to a Subsystem Manager; he or she ensures that the subsystem specifications are appropriated for achieving the project's goals,Adm,,
```
timj: From my dictionary app:
> The spellings adviser and advisor are both correct. Adviser is more common, but advisor is also widely used, especially in North America. Adviser may be seen as less formal, while advisor often suggests an official position.

So I think we leave it as "advisor".

womullan: ok

womullan: but https://www.merriam-webster.com/words-at-play/advisor-vs-adviser-who-will-win  
Adviser is the hands-down overall favorite, based on large swaths of data available on the Internet.
(and my US spell checker)

PR #189, opened 2019-07-28
Scan aglossry.tex until it converges

womullan:


womullan:
@jdswinbank this is not the niceest way to do it but certainly easiest and removes the uncertain number of calls in the makefile.

jdswinbank:
Nice – will try this out shortly.

bin/generateAcronyms.py
```diff
@@ -429,6 +428,24 @@ def write_latex_table(acronyms, fd=sys.stdout):
     print(r"\end{longtable}", file=fd)
 
 
+def forceConverge(incount, doGlossary, utags):
+    """Run throug the glossary looking for defnitions untill
```
jdswinbank: "through".

bin/generateAcronyms.py
```diff
@@ -429,6 +428,24 @@ def write_latex_table(acronyms, fd=sys.stdout):
     print(r"\end{longtable}", file=fd)
 
 
+def forceConverge(incount, doGlossary, utags):
```
jdswinbank: I'm not sure what you gain by passing `doGlossary` here — if you are calling `forceConverge` then `doGlossary` *must* be true, or you have already lost. I would pull this check outside the function.

bin/generateAcronyms.py
```diff
@@ -429,6 +428,24 @@ def write_latex_table(acronyms, fd=sys.stdout):
...
+    """Run throug the glossary looking for defnitions untill
+    no more are added.
+    """
+    if (not doGlossary):
```
jdswinbank: Here and below, there are a lot more parens than you actually need.

jdswinbank:
PR: https://github.com/lsst/lsst-texmf/pull/190

bin/generateAcronyms.py
womullan: since i started with LISP .. I feel i always am short of parens ..

PR #196, opened 2019-10-01
Add pyyaml dependency

SimonKrughoff:


jonathansick:
I'd recommend creating a `requirements.txt` file that collects all dependencies required by the various Python tools in `bin/` and then installing through that in the Docker file. Thus far the authors of `db2author.py` have been a bit reckless by requiring a third-party dependency and providing no canonical mechanism or documentation for a user to install it.

timj:
Thank you for calling me reckless but I never had any intention for this script to be run by everyone every time a document was built. My intent was for it to be run only by me and a few others and only towards the end of paper development and then using their science pipelines conda env which had all this available.

jonathansick:
Sorry about that, you're right. I'll change my comment.

SimonKrughoff:
OK, I made the suggested changes, but the docker build is not seeing the `requirements.txt`.  @jonathansick am I missing something?  Build is [here](https://travis-ci.org/lsst/lsst-texmf/jobs/592692057).

jonathansick:
You probably want to `pip install -r requirements.txt` *after* running `COPY` to get the lsst-texmf content into the image. (Or you could pre-load the requirements.txt in a docker build stage and run the pip installation then; I don't think it will make a meaningful efficiency different though)

SimonKrughoff:
Oh, of course.  Thanks for the pointer!

Dockerfile
```diff
@@ -11,6 +11,7 @@ MAINTAINER LSST SQuaRE <sqre-admin@lists.lsst.org>
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         csh \
+        python-pip \
```
timj: Isn't this a python2 pip? https://packages.ubuntu.com/bionic/all/python-pip/filelist

timj: I think you want https://packages.ubuntu.com/bionic/python3-pip

SimonKrughoff:
I have squashed my various attempts at this to a single commit.  I think this is now ready for review.

requirements.txt
```diff
@@ -0,0 +1 @@
+pyyaml
```
jonathansick: Do you think this should be pinned so we don't run into API changes from pyyaml?

SimonKrughoff: I'm content that we are using APIs that won't change in the foreseeable future.

PR #204, opened 2019-11-20
Correct the syntax of the "NVME" entry

gpdf:
Typo of '.' for ','


jdswinbank:
Isn't this term more commonly written as “NVMe”? (See e.g. https://nvmexpress.org, https://en.m.wikipedia.org/wiki/NVM_Express, etc). Your call if you want to fix that on this PR.

Otherwise, fine.

gpdf:
@jdswinbank good point, I'll add that change as soon as my rebase has passed the checks

PR #208, opened 2019-12-09
Add APDB and PPDB to glossary

fritzm:


etc/glossarydefs.csv
```diff
@@ -15,6 +15,7 @@
 "AIT","Assembly Integration and Test","Gen","",""
 "AIV","Assembly Integration and Verification","Gen","",""
 "ALMA","Atacama Large Millimeter Array (ESO)","Gen","",""
+"APDB", "Alert Product DataBase. A dedicated, internal database system used to support LSST Alert Production.  Does not support end-user access.","DM","",""
```
timj: Is it Product or Products?

jdswinbank: I have always assumed it was “Production”, to be consistent with our other uses of AP.

fritzm:
Yes, I believe it should be “Production”, but I typo’d.  Update coming...

PR #209, opened 2019-12-10
 ops proposal

womullan:
bib entry

### DIFF
--- a/texmf/bibtex/bib/lsst.bib
+++ b/texmf/bibtex/bib/lsst.bib
@@ -1380,22 +1380,6 @@
   note =         {Superseded by LDO-13},
 }
 
-@DocuShare{LDO-13,
-  author =      {R. Blum and others},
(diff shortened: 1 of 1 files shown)


texmf/bibtex/bib/lsst.bib
```diff
@@ -1380,6 +1380,14 @@ @DocuShare{LDO-13
   handle =      {LDO-13},
 }
 
+@DocuShare{LDO-31,
```
timj: My only comment is that it's now obvious that both these LDO documents should be further down the file after LSE-400

womullan: ok i can do that ... but whats the logic ?

timj: Mostly because grouping documents in the file by handle makes it less likely people will add the same document multiple times.

PR #210, opened 2019-12-10
If not glossary, only acronyms included in table

gcomoretto:
I opted for reducing the list of candidate acronyms to the (already defined) capitalized ones.

I think the implementation is very simple and is not interfering with the glossary extraction logic.


bin/generateAcronyms.py
```diff
@@ -492,6 +496,8 @@ def main(texfiles, doGlossary, utags):
 
     # Report missing definitions, taking into account skips
     missing = missing - skip
+    if not doGlossary:
```
timj: I think this is necessary because find_matches is still doing a broader search for acronyms and historically we did have some acronyms (e.g. "A/D") that really were acronyms but weren't just capital letters. I think we now filter those out?  I'm not entirely sure that merging acronyms and glossaries in the same file was right if we are now trying to split things up inside the code.

gcomoretto: We should improve the regular expression defining CAP_ACRONYM to include all edge-cases.

jdswinbank: I agree with / would go further than @timj here.

Specifically, “acronym” can mean almost anything. For example, look at [Wikipedia](https://en.wikipedia.org/wiki/Acronym): its very first examples of acronyms are “NATO“, “scuba“, and “Benelux”. Tim mentions “A/D”, and on #204 we also have “NVMe”. If you can make a regular expression which reliably captures all of those but drops other terms, you... probably have way too much time on your hands.

If we think distinguishing glossary entries from acronyms and abbreviations is a useful exercise, then I think it means keeping them in separate files (as Tim indicated). Just using “is presented in block capitals” as a heuristic to distinguish them seems too naïve to be useful.

gcomoretto: I have expanded the regular expression defining the variable CAP_ACRONYM, in order to:
- include numbers, but exclude string with only numbers
- include **/**, but not in the first position

Now, If we look at the list of defined acronyms in _etc/glossarydefs.csv_, we found **CFHT-LS**, that seems to be clearly an acronym, but will not be included in the acronyms table.

However, If we add **-** to the regex, this will pull out also all requirements, documents, RFCs, Jira issues and probably more.

We shall probably define what is meant to be an acronym in the LSST documentation (if not already done, maybe @jdswinbank  knows this).
Is  **-** a valid character to use in an acronym?

However, clearly now we are missing a few acronyms, defined already, like document handlers, that will never appear on its own.
For example, in the case of **LDM**-294, LDM is defined as an acronym in glossarydefs.csv but it will never be shown in glossary nor in the acronyms table.
This seems to me to be a different issue and requires probably additional logic to be added to the script.

Are there other special characters, like **.**, **\**, etc, that should be included in the regular expression?

jdswinbank: The only applicable project standard that I'm aware of is Document-13016. It does not define an acronym, but it gives the exemplars “km” and ”SNe”, so I think we can assume that they must be matched by any regexp defined here.

gcomoretto: OK, so **-** a valid character to use for acronyms?

What about the acronyms that we are missing?

jdswinbank: Sorry, perhaps I am not being clear above: I think it is impossible to write a regular expression that does what you want it to. Given that, I don't think discussing special characters is helpful.

gcomoretto: OK.
So, if it is fine to miss acronyms in the acronym table of in the glossary.

timj:
What I'm not sure about is why glossaries and acronyms were combined into a single file and we lost any ability to split them up again.  I agree that if we are going to define "acronym" as upper case letters then things are going to get inconsistent. The code already pulls out non-standard acronyms and handles them explicitly (which I think is why the glossary code works fine). @womullan  are you against either splitting the input files up or else putting an acronym flag in the CSV?

womullan:
I assumed you either use glossary or acronyms - if you use glossary that includes acronyms and the script differentiates Acronyms from Glossary entries as it writes the file.. A flag in the CSV would actually make this much cleaner and solve the other problem as well of just making an acronym file if i run in acronym mode.

gcomoretto:
OK. I will propose a different implementation based on identifying the acronyms in the CSV files with a flag.

gcomoretto:
I just pushed some changes that should complete the implementation.

A few additional changes have been done in order to take advantage of the new information provided in the CSV file. For example, when writing the glossary, the script will rely in the CSV characterization.

I have also reviewed the CSV ensuring that all acronyms have an **acronyms-like** definition (short), and created a separated entry for the corresponding glossary. We need to enforce this when reviewing future additions to the CSV.

Testing the script on different documents, plenty of missing acronyms found.
A different branch and PR will be used for that.

bin/generateAcronyms.py
```diff
@@ -39,7 +39,10 @@
 #  Match for extracting acronyms from the glossary myacronyms .txt files
 MATCH_ACRONYM = r"^([\w/&\-\+ -']+)\s*:\s*(.*)$"
 MATCH_ACRONYM_RE = re.compile(MATCH_ACRONYM)
-CAP_ACRONYM = re.compile(r"\b[A-Z][A-Z]+\b")
+# following regular expression define an acronym-like string as follows:
+# - including upper case characters and numbers
+# - starting with a letter
+CAP_ACRONYM = re.compile(r"\b(?!\d+\b)[A-Z][A-Z0-9]+\b")
```
timj: What is the `(?!\d+\b)` really doing here?  I don't think it is changing the match at all because without it it still requires word boundary then a capital letter.

bin/generateAcronyms.py
```diff
@@ -492,6 +498,10 @@ def main(texfiles, doGlossary, utags):
 
     # Report missing definitions, taking into account skips
     missing = missing - skip
+
+    print("List of potential acronyms found in your text"
```
womullan: This is missing some space and ) -- this is the output on execution 
"List of potential acronyms found in your textthat you may be missing. Please ignore if not relevant (note that the list may not be complete."

Also it is output even when there are NO missing acronyms listed. So if m is empty this should e suppressed.

gcomoretto:
I have pushed a change to resolve Wil's comment.

bin/generateAcronyms.py
```diff
@@ -112,14 +115,17 @@ def read_glossarydef(filename, utags, init=None):
...
                 continue  # There is a header line
             ind = 0
             acr = row[ind]
-            ind = ind + 1
-            defn = row[ind]
-            ind = ind + 1
-            tags = row[ind]
+            defn = row[ind + 1]
+            tags = row[ind + 2]
+            type = row[ind + 5]
```
timj: Don't use a `type` variable name since it clashes with the builtin. Use `entryType` or something.

bin/generateAcronyms.py
```diff
@@ -331,6 +337,7 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
     # Also single character acronyms (which should probably be banned)
     nonstandard = {a for a in acronyms if not a.isupper() or not
                    a.isalpha() or len(a) == 1}
+    # CAP_ACRONYM cannot be used here since it search for substrings
```
timj: Move this before the code it is referring to.

bin/generateAcronyms.py
```diff
@@ -492,8 +498,13 @@ def main(texfiles, doGlossary, utags):
 
     # Report missing definitions, taking into account skips
     missing = missing - skip
-    for m in missing:
-        print("Missing definition: {}".format(m), file=sys.stderr)
+
+    if len(missing) > 0:
```
timj: You can drop the `> 0` since `len(missing)` is false if the set is empty and true otherwise.

bin/generateAcronyms.py
gcomoretto: I added this when I updated the regex to include upper case characters and numbers, also in the first position. This avoid having string of only digits.
Now that the regex has been modified to start with an upper case character, this is not required anymore. I remove it.

PR #216, opened 2020-01-08
Add Dino Bektesevic to the list of authors in authordb.yaml.

DinoBektesevic:
Added myself into authorsdb.yaml file. Required to finalize https://github.com/lsst-dm/dmtn-137. 

Had to fork and pull again because I don't have sufficient permissions to work with lsst repos directly.


DinoBektesevic:
Thanks! I do need you to merge it. Only when i'm added to the git repo and get push and merge permissions can I merge it - but in that case I would not have had to fork it in the first place.

womullan:
done

PR #219, opened 2020-01-11
update the header printed to authors.tex by db2authors.py

bxin:


bin/db2authors.py
```diff
@@ -71,7 +71,7 @@
 
 print("""%% DO NOT EDIT THIS FILE. IT IS GENERATED FROM db2authors.py"
 %% Regenerate using:
-%%    python support/db2authors.py > authors.tex
+%%    python $LSSTTEX/bin/db2authors.py > authors.tex
```
timj: `$LSSTTEX` isn't a variable that people set. People do set TEXMFHOME so one option here is `$TEXMFHOME/../bin/` for the path. Another is to be a little more descriptive and say `$LSST_TEXMF_DIR/bin` (which also is not a variable we set but is obvious what we mean).

bxin: Thanks. $LSST_TEXMF_DIR was what was intended.

PR #220, opened 2020-01-13
updated glossary/acronyms for LSST and VRO

MelissaGraham:
Changes:
"LSST","Legacy Survey of Space and Time (formerly Large Synoptic Survey Telescope)","Gen","","","A"

Additions:
"VRO","Vera C. Rubin Observatory","Gen","","","A"

Did not add because it's unclear this is an acronym going to be used, and because SST is already "Subsystem Science Team":
"SST","Simonyi Survey Telescope","Gen","","","A"


timj:
How about adding Simonyi Survey Telescope to the glossary part though? "The telescope that will perform the LSST" or something similar?

MelissaGraham:
> How about adding Simonyi Survey Telescope to the glossary part though? "The telescope that will perform the LSST" or something similar?

Good idea... I'll go with "The telescope at the VRO that will perform the LSST (this refers to all physical components: the mirror, the mount assembly, etc.)."

etc/glossarydefs.csv
```diff
@@ -312,7 +314,8 @@
 "DMSE","Data Management System Engineer","LSST DM","","","A"
 "DMSR","DM System Requirements; LSE-61","LSST DM","","","A"
 "DMSS","DM Subsystem Scientist","LSST DM","","","A"
-"DMSST","DM System Science Team","LSST DM","","","A"
+"DM-SST","Data Management Subsystem Science Team","LSST DM","","DMSST","A"
+"DMSST","DM System Science Team (see DM-SST)","LSST DM","","DM-SST","A"
```
jdswinbank: I'm not sure if the above is saying that “DM-SST” and “DMSST” are the same or different! :-)

For what it's worth, the official name of the team is the “DM System Science Team” (ref LDM-294 §7.1). I have no position on whether or not the abbreviation should contain a hyphen!

MelissaGraham: Oh whoops I didn't notice, the "Subsystem" vs. "System" usage snuck in on me. But it should be "System" not "Subsystem". I'm going to fix this so that "DM-SST" and "DMSST" have *exactly* the same definition: "DM System Science Team".

womullan:
we are not supposed to use VRO - i have rebased and noted this in the entry.

PR #221, opened 2020-01-14
Clarify the difference between "SUI" and "SUIT"

gpdf:


jdswinbank:
These definitions look fine.

However, I think (but am not aware of any written guidance to this effect) that we conventional limit the definition of the acronym to just its direct expansion, and then provide a glossary entry which provides further information (rather than using parentheses, as you've done here). See e.g. [DRP](https://github.com/lsst/lsst-texmf/blob/0f109838427dde5a1c8d5ffdee85ce017aad63e4/etc/glossarydefs.csv#L320) and [Data Release Production](https://github.com/lsst/lsst-texmf/blob/0f109838427dde5a1c8d5ffdee85ce017aad63e4/etc/glossarydefs.csv#L576).

gpdf:
Makes sense; I'll fix that.  I see that there is a pattern of putting the acronym definition immediately adjacent to the glossary entry in the file.

Two more questions after browsing through the file for a bit:

1. I see that the file has several sub-sections each in alphabetical order.  I can make sense of some of the divisions, but not all of them.  Is there a clear definition for each block?
1. I see that "HiPS" is mis-capitalized "HIPS" in the file.  If I fix this will this break existing links in documents that used the acronym?  Is the matching case-sensitive or not?

womullan:
The glossary was sorted once upon a time .. after that things were added .  I would leave the HIPS .. though it might/should be ok to change .. also depends on whether its used as acronym or glossary.

jdswinbank:
Given @womullan's comment, I'd suggest that we simply sort the file. @gpdf , do you want to do that on this PR?

I don't understand Wil's comment about HiPS. I agree that HiPS is the correct capitalization. I understood that @gcomoretto added the `A` and `G` notation precisely so that abbreviations don't have to be in block capitals. Given that, I don't see any reason not to get this right... but perhaps that's because I don't fully understand the situation.

timj:
HiPS should work fine.

womullan:
Great then lets sort the file - and merge it quickly !

womullan:
I have now re sorted the file. You should pu tn these changes but it may be better to do them afresh ..

PR #232, opened 2020-02-19
U/womullan/tabacronyms

womullan:


womullan:
I have not sorted the acronyms yet .. 
but I did change gnerateAcronyms non distructively

bin/generateAcronyms.py
```diff
@@ -18,6 +18,8 @@
 
 Passing ``-u`` or ``--update`` will process all LaTeX files to identify
 potential glossary entries and will mark them with ``\gls``.
+
+Passing ``-n`` or ``-notex`` will produce a plain text tab delimeted list.
```
timj: delimeted -> delimited

Also probably `--notex` and not `-notex`

bin/generateAcronyms.py
```diff
@@ -409,7 +411,7 @@ def write_latex_glossary(acronyms, fd=sys.stdout):
                       acr, acr, defn[0]), file=fd)
 
 
-def write_latex_table(acronyms, fd=sys.stdout):
+def write_latex_table(acronyms, notex=False, fd=sys.stdout):
```
timj: I usually prefer booleans to have a positive sense because it's easier to reason out what will happen. Negative booleans are hard.  Consider `plaintext=False` instead, or `latex=True`.

bin/generateAcronyms.py
```diff
@@ -522,12 +531,16 @@ def main(texfiles, doGlossary, utags):
...
+        acrFile = "acronyms.txt"
+
     if doGlossary:
         with open(glsFile, "w") as gfd:
             write_latex_glossary(results, fd=gfd)
     else:
-        with open("acronyms.tex", "w") as fd:
-            write_latex_table(results, fd=fd)
+        with open(acrFile, "w") as fd:
+            write_latex_table(results, notex, fd=fd )
```
timj: Remove that space before the final bracket.

bin/generateAcronyms.py
```diff
@@ -419,7 +421,10 @@ def write_latex_table(acronyms, fd=sys.stdout):
     acronyms : `list`
         List of 2-tuples with acronym and definition.
     """
-    print(r"""\addtocounter{table}{-1}
+    if (notex):
+       print("Acronym\tDescription", file=fd)
```
timj: Indent is wrong here -- we use four for python. Also in the `else`. It's why travis is failing.

bin/generateAcronyms.py
```diff
@@ -419,7 +421,10 @@ def write_latex_table(acronyms, fd=sys.stdout):
     acronyms : `list`
         List of 2-tuples with acronym and definition.
     """
-    print(r"""\addtocounter{table}{-1}
+    if (notex):
```
timj: Don't put parens around `notex`.

bin/generateAcronyms.py
```diff
@@ -428,9 +433,13 @@ def write_latex_table(acronyms, fd=sys.stdout):
         acr = acr.replace("&", r"\&")
         acr = acr.replace("_", r"\_")
         defn = glsreg.sub(glsrmfn, defn[0])
-        print("{} & {} {}".format(acr, defn, r"\\\hline"), file=fd)
+        if (notex):
```
timj: If you wanted to stop repeating the print you could consider:
```python
if notex:
    sep = "\t"
    end = ""
else:
    sep = " & "
    end = r" \\\hline"
print(f"{acr}{sep}{defn}{end}", file=fd)
```

bin/generateAcronyms.py
```diff
@@ -428,9 +433,13 @@ def write_latex_table(acronyms, fd=sys.stdout):
...
+            print("{}\t{}".format(acr, defn), file=fd)
+        else:
+            print("{} & {} {}".format(acr, defn, r"\\\hline"), file=fd)
 
-    print(r"\end{longtable}", file=fd)
+    if (not notex):
```
timj: drop the parens.

etc/glossarydefs.csv
```diff
@@ -105,6 +117,7 @@
 "GAVO","German Astronomical Virtual Observatory","Gen","","","A"
 "Gb","Gigabit","Gen","","","A"
 "GB","Gigabyte","Gen","","","A"
+"GEANT","pan-European data network for the research and education community","Gen","","","A"
```
jdswinbank: If it's an acronym, we should give the expansion, not just a meaning.

etc/glossarydefs.csv
```diff
@@ -186,16 +202,23 @@
 "NASA","National Aeronautics and Space Administration","Gen","","","A"
 "NED","NASA/IPAC Extragalactic Database","Gen","","","A"
 "NCOA","National Center for Optical-Infrared Astronomy","Gen","","","A"
+"NMOC","NSF's OIR Lab Management Oversight Council","Gen","","","A"
 "NSF's OIR Lab","\gls{NSF}'s National Optical-Infrared Astronomy Research Laboratory; \url{https://nationalastro.org}","Gen","","","A"
 "NCSA","National Center for Supercomputing Applications","Gen","","","A"
 "NEO","Near-Earth Object","Gen","","","A"
 "NFS","Network File System","Gen","","","A"
+"NIR","Near Infra Red","Sci","","","A"
```
jdswinbank: Infrared is one word

etc/glossarydefs.csv
```diff
@@ -186,16 +202,23 @@
...
+"NUV","Near Ultra Violet","Sci","","","A"
```
jdswinbank: as is ultraviolet

etc/glossarydefs.csv
```diff
@@ -271,11 +305,15 @@
...
 "VO","Virtual Observatory","Gen","","","A"
 "VOIP","Voice Over Internet Protocol","IT DM","","","A"
+"VST","VLT Survey Telescope","Gen","","","A"
 "WAN","Wide Area Network","Gen","","","A"
 "WBS","\gls{Work Breakdown Structure}","Gen","","","A"
 "WCS","\gls{World Coordinate System}","Gen","","","A"
+"WFD","Wide Fast Deep","OPS","","","A"
 "WISE","Wide-field Survey Explorer","Gen","","","A"
 "WSDL","Web Services Description Language","Gen","","","A"
+"WL","Weak gravitational Lens cosmic shear","Sci","","","A"
```
jdswinbank: Just “weak lensing” is better.

bin/generateAcronyms.py
```diff
@@ -522,12 +531,16 @@ def main(texfiles, doGlossary, utags):
         else:
             raise RuntimeError("Internal error handling {}".format(acr))
 
+    acrFile = "acronyms.tex"
```
timj: consider:
```python
ext = ".txt" if notex else ".tex"
acrFile = f"acronyms.{ext}"
```

etc/glossarydefs.csv
```diff
@@ -255,13 +284,18 @@
...
 "UI","User Interface","Gen","","","A"
+"URL","Universal Resource Locator","Gen","","","A"
+"UKIDSS","UKIRT Infrared Deep Sky Survey","Gen","","","A"
 "US","United States","Gen","","","A"
+"USDF","United States Data Facility","OPS DF DM","","","A"
 "USNO","United States Naval Observatory","Gen","","","A"
 "UT","Universal Time","Gen","","","A"
 "UT1","Universal Time 1","Gen","","","A"
 "UTC","Coordinated Universal Time","Gen","","","A"
+"UKIRT","United Kingdom Infra-Red Telescope","Gen","","","A"
```
timj: Infrared

womullan: copied from their web page I thought .. but ok

bin/generateAcronyms.py
womullan: perhaps later ..   now it all works ..

etc/glossarydefs.csv
timj: I did work there for 15 years. I imagine that since UKIRT was sold that it probably doesn't stand for anything at all now.

bin/generateAcronyms.py
womullan: it appears i fixed 3 flake errors which were not pushed .. including this

bin/generateAcronyms.py
womullan: did similar

bin/generateAcronyms.py

etc/glossarydefs.csv
womullan: .. ok so perhpas its not an acronym ..  i can flip to G

etc/glossarydefs.csv
womullan: ok

bin/generateAcronyms.py
womullan: done

womullan:
File is now sorted also

PR #233, opened 2020-02-22
Tickets/rhl/rubin templates

womullan:
Thanks to Robert for updating this - i have tried it out and tweaked the default.


bin/showBeamerThemes
```diff
@@ -0,0 +1,98 @@
...
+                    break
+
+    if themeDir is None:
+        print("Unable to find LSST-themes on your TeX search path")
+        sys.exit(1)
+
+    #-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-
+
+    print(r"""
+    \documentclass[12pt, t, aspectratio=%(aspectratio)s, %(xetex)s]{beamer}
```
timj: Maybe use an f-string here for clarity.

timj:
Travis is failing with:
```
LaTeX Error: File `LSST-themes/Rubin/LSSToptions.tex' not found
```

bin/showBeamerThemes
womullan: noted on the document template we can do that another day .. will look at script now

PR #237, opened 2020-03-09
-s to allow SPIE authors

womullan:


bin/db2authors.py
```diff
@@ -32,6 +33,20 @@
 # the current working directory.
 authorfile = os.path.join("authors.yaml")
 
+description = __doc__
+formatter = argparse.RawDescriptionHelpFormatter
+parser = argparse.ArgumentParser(description=description,
+                                 formatter_class=formatter)
+
+parser.add_argument('-s', '--spie', action='store_true',
```
timj: I think if this is a trend to supporting different variants then I'd prefer a `--mode=spie` option where `aas` is the default. See eg https://github.com/lsst/astro_metadata_translator/blob/master/python/astro_metadata_translator/bin/translateheader.py#L113

bin/db2authors.py
```diff
@@ -32,6 +33,20 @@
...
+                    help=""" iOutput suitable for SPIE paper.""")
+args = parser.parse_args()
+
+cmd = "affiliation"
```
timj: Maybe `affil_cmd` as the variable? `cmd` seems too generic.

womullan: ok

PR #240, opened 2020-03-14
RST output for acronyms

womullan:


womullan:
Switched to mode - I only yuse -t in one place and its the conversation we had about db2authors - mode is a more flexible and correct way to do this now there are 3 outputs

etc/authordb.yaml
```diff
@@ -29,7 +29,7 @@ affiliations:
   Duke: Department of Physics, Duke University, Durham, NC 27708, USA
   EPFL: Institute of Physics, Laboratory of Astrophysics, \'{E}cole Polytechnique
     Fed\`{e}rale de Lausanne (EPFL), Observatoire de Sauverny, 1290 Versoix, Switzerland
-  Facebook: Facebook,  1 Hacker Way, Menlo Park, California, USA
+  Facebook: Facebook,  14734 Friend Plaza, Springfield, NE 68059, USA
```
timj: I think I'd probably call this FacebookNE because it's not the standard location for Facebook (we'd need to change the author entry to match as well).

womullan: ok

jonathansick:
Sorry I couldn't get to the review over the weekend. Having looked at this now, I'm fine with this functionality being prototyped within lsst-texmf to serve immediate needs.

To be honest, I don't think this is the ideal long-term acronym and glossary solution for reStructuredText/Sphinx-based technotes and documentation though.

I see lsst-texmf as a data source for acronyms, glossary terms, and authorship information (as it already is for bibliography terms), but the code (implemented differently, though) should live in [Documenteer](https://documenteer.lsst.io) with the rest of DM's Sphinx extensions.

I'll reprioritize adding that functionality to Documenteer so we don't run into confusion about how acronyms and glossaries need to be handled in Sphinx projects.

womullan:
not sure why you would waste time rewriting this - it was not there before so no point adding it now if you don't like the implementation I have given you. I actually also dislike the copy of some resources into the RST docs we should fix this with a common submodule or somethign like that. Or perhaps all of that code and resources belong in documenteer. But implementing the same functionality in another place differently makes no sense.

PR #247, opened 2020-04-09
works for SPIE with numerals for superscript

womullan:


gcomoretto:
It seems it works for me, I have fixed the Makefile in DMTN-140, and authors.tex removed from the repository

bin/db2authors.py
```diff
@@ -75,6 +76,7 @@
 # dict of all the affiliations, key is a label
 # used in author list
 affil = authordb["affiliations"]
+affilset = list()
```
timj, line 79: It's confusing to me to have "set" in the variable name associated with a `list`.

bin/db2authors.py
```diff
@@ -103,8 +106,25 @@
     except KeyError as e:
         raise RuntimeError(f"Author ID {authorid} now defined in author database.") from e
 
-    if "orcid" in auth and auth["orcid"]:
-        orcid = "[{}]".format(auth["orcid"])
+    affilOutput = list()
+    affilAuth = ""
+    affilSep = ""
+    for theAffil in auth["affil"]:
+        if (theAffil not in affilset):
```
timj, line 113: You don't need the brackets here.

bin/db2authors.py
womullan: well i would use set but it does not support index .. it is a set as i do not allow suplicates .unlike a list ..

PR #252, opened 2020-04-24
generic tool to pull tables from google sheet to tex table

womullan:


bin/makeTablesFromGoogle.py
```diff
@@ -0,0 +1,197 @@
...
+        if (i > 0 and skipped < skip):
+            skipped = skipped + 1
+        else:
+            try:
+                print("%s{%s}" % (pre, fixTex(row[i])), end='', file=tout)
+            except IndexError:
+                pass
+            if i < cols-1:
+                print("&", end='', file=tout)
+    print(" \\\\ \\hline", file=tout)
```
timj, line 107: might be more readable using r-strings here.

bin/makeTablesFromGoogle.py
```diff
@@ -0,0 +1,197 @@
...
+def fixTex(text):
+    ret = text.replace("_", "\\_")
+    ret = ret.replace("/", "/ ")
+    ret = ret.replace("$", "\\$")
+    ret = ret.replace("%", "\\%")
+    ret = ret.replace("^", "\\^")
+    return ret
+
+
+def genTables(values):
```
timj, line 119: It took me a while to disentangle what this function was doing so I think it might be worth a short docstring here saying that files are opened as each new table is encountered and there is some kind of hint in the spreadsheet to indicate when that is. It would also help readability if `outtail` was renamed to something like `complete_and_close`.

bin/makeTablesFromGoogle.py
```diff
@@ -0,0 +1,197 @@
...
+    name = ''
+    tout = ''
+
+    if not values:
+        print('No data found.')
```
timj, line 124: If you return here you don't have to worry about outtail being called later with tout='' and you can get rid of one level of indent.

bin/makeTablesFromGoogle.py
```diff
@@ -0,0 +1,197 @@
...
+    ret = ret.replace("/", "/ ")
+    ret = ret.replace("$", "\\$")
+    ret = ret.replace("%", "\\%")
+    ret = ret.replace("^", "\\^")
+    return ret
+
+
+def genTables(values):
+    name = ''
+    tout = ''
```
timj, line 121: `None` is a more normal default for a file handle.

bin/makeTablesFromGoogle.py
```diff
@@ -0,0 +1,197 @@
...
+                if name and row:
+                    if (row[0].startswith('Year')
+                            or row[0].startswith('Total')
+                            or bold_next):
+                        # print header/total in bold
+                        outputrow(tout, "\\textbf", row, cols, skip)
+                        bold_next = False
+                    else:
+                        outputrow(tout, "", row, cols, skip)
+    outtail(tout)
```
timj, line 154: Shouldn't this do `if tout:` to ensure that an open file is going to be closed?

bin/makeTablesFromGoogle.py
womullan: ok

bin/makeTablesFromGoogle.py
womullan: added docstring
changed name  complete_and_close_table

bin/makeTablesFromGoogle.py
womullan: got a real problem if there is now open file.
complete_and_close_table shoudl really check this not each call.  I put an exception in there

bin/makeTablesFromGoogle.py
womullan: ok

PR #255, opened 2020-05-05
DM-23570: Switch document look and feel to Rubin Observatory

womullan:


texmf/ls-R
```diff
@@ -396,6 +396,7 @@ LSSToptions.tex
 lsst
 
 ./tex/latex/lsst:
+.lsstdoc.cls.swp
```
timj, line 399: I don't think this should be here.

texmf/tex/latex/lsst/lsstdoc.cls
```diff
@@ -474,24 +455,24 @@
...
+   The control is on the most recent digital document with this Handle in the Rubin digital archive and not printed versions.
    }
    \def\@controlFootText{%
    The contents of this document are subject to configuration control and may not be changed, altered, or their provisions waived without prior approval.
    }
    \fi
    \if@dmccb
    \def\@controlText{%
-   This LSST document has been approved as a Content-Controlled Document by the LSST DM Change Control Board.
+   This Rubin document has been approved as a Content-Controlled Document by the Rubin DM Change Control Board.
```
timj, line 469: I think we probably want to be saying "Rubin Observatory" here whenever "Rubin" is used.

texmf/tex/latex/lsst/lsstdoc.cls
```diff
@@ -722,7 +703,7 @@
 \setlength{\lsstheadlogooffset}{\headextrawidth}
 
 % Text to use below the logo in the header
-\newcommand{\headertext}{\tiny LARGE SYNOPTIC SURVEY TELESCOPE}
+\newcommand{\headertext}{\tiny Vera C. Rubin Observatory}
```
timj, line 706: You might want to do a quick test with pdflatex to see if that hard coded length below needs to be tweaked.

texmf/ls-R
womullan: should not be

texmf/tex/latex/lsst/lsstdoc.cls
womullan: ok

texmf/tex/latex/lsst/lsstdoc.cls
womullan: currently not used since the logo is "Vera C Rubin Observatory" - when we get a log it may come back ..

PR #282, opened 2020-08-10
glossary check  and nicer(?) glossary javascript page

womullan:
Perhaps the try except could be wider - plus @timj may not like that i catch report and throw the exception on ... I guess I could stuff the message into a new exception ..

### DIFF
--- a/bin/generateAcronyms.py
+++ b/bin/generateAcronyms.py
@@ -127,8 +127,7 @@ def read_glossarydef(filename, utags, init=None):
                 tags = row[ind + 2]
                 entryType = row[ind + 5]
             except BaseException as ex:
-                print("Error reading {} on line {} - {}".format(filename, lc, row),
-                      sys.exc_info()[0])
+                print("Error reading {} on line {} - {}".format(filename, lc, row))
                 raise ex
 
             if not doGlossary and entryType == "G":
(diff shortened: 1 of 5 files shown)


bin/generateAcronyms.py
```diff
@@ -122,10 +121,16 @@ def read_glossarydef(filename, utags, init=None):
...
-            tags = row[ind + 2]
-            entryType = row[ind + 5]
+            try:
+                acr = row[ind]
+                defn = row[ind + 1]
+                tags = row[ind + 2]
+                entryType = row[ind + 5]
+            except BaseException as ex:
+                print("Error reading {} on line {} - {}".format(filename, lc, row),
+                      sys.exc_info()[0])
```
timj, line 131: I'm not sure what the intent is here? Doesn't this print something like "<class 'IndexError'>" ?

I think if you want the class name reported you can do that with `type(ex).__name__`. You will already get the IndexError from reraising in the next line. Maybe this was a debug leftover?

bin/generateAcronyms.py
```diff
@@ -465,16 +470,21 @@ def forceConverge(prevCount, utags):
         prevCount = count
 
 
+def setup_paths():
+    defaults_dir = os.path.join(
+        os.path.dirname(__file__), os.path.pardir, "etc")
+    lsst_glossary_path = os.path.join(defaults_dir, "glossarydefs.csv")
+    global_skip_path = os.path.join(defaults_dir, "skipacronyms.txt")
+    return [lsst_glossary_path, global_skip_path]
```
timj, line 478: Return a tuple here and not a list.

bin/generateAcronyms.py
womullan: in deed it prints -- Error reading /Users/womullan/LSSTgit/lsst-texmf/bin/../etc/glossarydefs.csv on line 11 - ['ACCS', 'Auxiliary Camera Control System,LSST DM"', '', '', 'A'] <class 'IndexError'>
I should remove the IndexError part all I wanted was the line  number and entry

bin/generateAcronyms.py

PR #298, opened 2020-10-02
now also accepts myglossarydefs.cvs as local defs

womullan:


bin/generateAcronyms.py
```diff
@@ -187,25 +195,31 @@ def read_myacronyms(filename="myacronyms.txt", allow_duplicates=False,
         Dictionary with acronym as key and definition as value.
     """
     definitions = {}
+    if filename.endswith(".csv"):
+        localdefs = read_glossarydef(filename, utags)
+        # flatten the set to a single entry for each glossarydef
+        for d in localdefs:
```
timj, line 201: I think you can do that with:
```python
definitions = {d: options.pop() for d, options in localdefs.items()}
```

docs/lsstdoc.rst
```diff
@@ -231,13 +231,21 @@ One should note particularly the Subsystem Tags which may be used to differentia
...
+   One line per abbreviation.
+
 :file:`myacronyms.txt`
    Defines abbreviations which are specific to this document.
    One line per abbreviation, formatted as ``Abbreviation:Definition`` (e.g. ``CI:Continuous Integration``).
 
-:file:`skipacronyms.txt`
-   Specifies abbreviations which should be omitted from the glossary.
-   One line per abbreviation.
+   **OR**
```
timj, line 242: I think you probably want to state that myglossary is only read is myacronyms is not read (ie myacronyms takes precedence and we don't merge the two file contents).

timj (approved):

docs/lsstdoc.rst
womullan: ok

bin/generateAcronyms.py
womullan: ok .. compound but works

PR #299, opened 2020-10-07
Fixes to issues that hang .tex compiler

rdepeyster:
These fixes address issues in glossarydefs.csv that hang .tex compiler when attempting to create PDF of all glossary records


timj:
Can you please separate the compilation fix into a distinct commit so it is separate from what look like general cleanups that are not part of the latex problem?

It looks like we do put latex directly in that file so the `\&` looks to be our only choice (we can't fix it in the python that reads it) -- @womullan I think I'd prefer it in general if we didn't write latex in what is meant to be a general purpose glossary definition but that's nothing to do with this PR.

womullan:
ok I have not liked the \gls stuff too much in there either. I guess one should process the entries in generate acronyms before writing them and add \gls to the acronym f the glossary entry exists. It would be more consistent and leave the data file as just data (also better for the web page version)

etc/glossarydefs.csv
```diff
@@ -812,7 +812,7 @@ WLMS,work load management service,TS,Document-11982,,A
 WMS,Work Management System,OPS,,,A
 WOUCAO,Windows on the Universe Center for Astronomy Outreach,OPS,,,A
 WP,Work Package,OPS,,,A
-WRHEN,Western Hemisphere Research & Education Networks,TS,Document-12033,,A
+WRHEN,Western Hemisphere Research \& Education Networks,TS,Document-12033,,A
```
womullan, line 815: i would put and or fix the script to escape these in the out put.

timj:
Yes, I agree that latex markup should be added later and we have the ability to add the gls ourselves as well. Not for this PR though but definitely something that should be added later (as you say, it would help the HTML view).

womullan:
https://jira.lsstcorp.org/browse/DM-26672 unless @rdepeyster  wants to have a go at it

etc/glossarydefs.csv
rdepeyster: Will, Tim - Thanks for the input. The glossardefs.csv changes I proposed are related to resolving issues that cause the GitHub .tex compiler to stop when attempting to PDF the entire glossary per RTN-004. None of the changes are general cleanup, though it might seem that way. For example, addition of acronym entries for ImSim and SLAC correspond to taking those acronyms out of the Term fields for Image Simulation and SLAC National Laboratory glossary entries (they had been in parenthesis in the Term fields, which caused compiler hangs). Spelling mismatches between some of the acronym entries and glossary items they referred to (via \gls) also caused hangs. And I removed one of the two entries for the ADCO acronym as the duplicate entries caused a hang (and, in this case, it didn't make sense to resolve by putting one of the two entries into myacronyms.txt). And the unescaped & hangs. So I'm not clear about what else I need to do to root the gotchas out of glossarydefs.csv. Please let me know the next step. Thanks!

timj:
@rdepeyster thanks for the clarification -- some of those changes looked like formatting fixes "test stand" vs "Test Stand" so I got confused.  Given the current state of the system I'm fine with you merging your fixes (rebase please, do not hit "update branch") and we can worry about how to sort out the latex that should not be in that file later on.

timj:
Actually, can you add a build of the glossary in latex to .github/workflows/ci.yaml so that we make sure no-one breaks it in the future?

womullan:
and we should link it to https://www.lsst.org/scientists/glossary-acronyms   if we can get it done automatically. How are you handling the multiple entries ? I assumed you would modify generateacronyms.py to make a table of the sorted contents of glossarydefs ..

womullan:
@rdepeyster please merge this - it will get very out of date otherwise.

etc/glossarydefs.csv
womullan: in my branch i have fixed the script to escape & and _  in definitions - was done for acronyms not glossary entries

PR #301, opened 2020-10-09
RPF

womullan:


etc/glossarydefs.csv
```diff
@@ -630,12 +630,13 @@ RENATER,Réseau National de télécommunications pour la Technologie l’Enseign
...
 REUNA,Red Universitaria Nacional,Gen,,,A
 RFC,Request For Comment,LSST DM,,,A
-RFP,Request For Proposals,LSST OPS,,,A
+RFP,Request For Proposals,PMO,,,A
 RFQ,Request For Quotations,LSST OPS,,,A
 RM,Release Manager,LSST DM,,,A
 RMS,Root-Mean-Square,Gen,,,A
 ROO,Rubin Observatory Operations,OPS,,,A
 ROOT,Object-oriented data analysis framework developed at CERN,Gen,,,A
+RPF,Rubin System PerFormance,OPS,,,A
```
leannep, line 639: Maybe add department Rubin System PerFormance department  to be clear that this an operations department name not a system performance analysis/

womullan: but we would have  to do it for all then RDP REO ROO ..

PR #304, opened 2020-10-17
update gloassary output to add \gls remove all \gls from defn file

womullan:
Also escapes & _ in all the out put these are also removed from glosarydef


bin/generateAcronyms.py
```diff
@@ -461,20 +463,19 @@ def write_latex_table(acronyms, dotex=True, dorst=False, fd=sys.stdout):
     for acr, defn in acronyms:
         acr = acr.replace("&", r"\&")
         acr = acr.replace("_", r"\_")
-        defn = glsreg.sub(glsrmfn, defn[0])
```
timj, line 464: Removing this requires that glsreg is removed as well from a couple of lines above.

bin/generateAcronyms.py
```diff
@@ -576,6 +579,27 @@ def main(texfiles, doGlossary, utags, dotex, dorst, mode):
     return len(results)
 
 
+def update_gls_entries(results, GLSlist):
+    """ Scan through the acronym anf gls definitions and add \gls where
```
timj, line 583: typo: anf

bin/generateAcronyms.py
```diff
@@ -593,10 +617,33 @@ def glsfn(s):
...
+    """ Make a re map of regexps for substitution"""
+    regexmap = {}
+    for g in GLSlist:
+        if g != '':
+            regexmap[g] = re.compile(r"([,\s(](?<!={))("+g+r")(\b)")
+    return regexmap
+
+def sub_line (line, regexmap, GLSlist):
+    """ for give line put \gls around gls items not already adorned.
+    :param line:
```
timj, line 630: Either use numpydoc following DM convention, or remove the `:param:` lines completely (as written they are just repeating the arguments anyhow).

bin/generateAcronyms.py
```diff
@@ -593,10 +617,33 @@ def glsfn(s):
...
 
-
-def glsrmfn(s):
-    """remove \\gls{} -- used in the regexp substitution for acronym"""
-    return s.group(1)
+def make_regexmap (GLSlist):
+    """ Make a re map of regexps for substitution"""
+    regexmap = {}
+    for g in GLSlist:
+        if g != '':
```
timj, line 624: Is there a reason that `if g:` doesn't work?

bin/generateAcronyms.py
```diff
@@ -593,10 +617,33 @@ def glsfn(s):
...
+            regexmap[g] = re.compile(r"([,\s(](?<!={))("+g+r")(\b)")
+    return regexmap
+
+def sub_line (line, regexmap, GLSlist):
+    """ for give line put \gls around gls items not already adorned.
```
timj, line 629: This needs to be an r-string because of that `\gls`.

give -> given

bin/generateAcronyms.py
```diff
@@ -593,10 +617,33 @@ def glsfn(s):
...
+    :param line:
+    :param regexmap:
+    :param GLSlist:
+    :return: modifed line
+    """
+    nline = line
+    for g in GLSlist:
+        regx = regexmap[g]
+        res = regx.search(line)
+        if (res is not None):  # ok now .. more checks
```
timj, line 639: The parentheses are not needed

bin/generateAcronyms.py
```diff
@@ -593,10 +617,33 @@ def glsfn(s):
...
+        if g != '':
+            regexmap[g] = re.compile(r"([,\s(](?<!={))("+g+r")(\b)")
+    return regexmap
+
+def sub_line (line, regexmap, GLSlist):
+    """ for give line put \gls around gls items not already adorned.
+    :param line:
+    :param regexmap:
+    :param GLSlist:
+    :return: modifed line
```
timj, line 633: modifed to modified

bin/generateAcronyms.py
```diff
@@ -605,25 +652,13 @@ def updateFile(inFile, GLSlist):
...
-                        res = regx.search(line)
-                        if (res is not None):  # ok now .. more checks
-                            # check its not a word in a GLS item but not
-                            # too greedy
-                            glsed = re.search(r"gls{.+"+g+"[a-z,A-Z, ]*}", line)
-                            if (glsed):  # already glsed or contained in one
-                                continue
-                            else:  # .. find and add GLS -
-                                line = regx.sub(glsfn, line)
+                    sub_line(line,regexmap,GLSlist)
```
timj, line 661: I'm a bit confused because this isn't returning anything so what is the sub_line meant to do?

timj (approved):
I don't understand what sub_line is doing in one place.

Lots of flake8 problems need fixing.

bin/generateAcronyms.py
womullan: ok .. its not removed only refactored ..

timj: My mistake. I thought the flake8 report said it was no longer used.

bin/generateAcronyms.py
womullan: ok

bin/generateAcronyms.py
womullan: ok

bin/generateAcronyms.py
womullan: ok

bin/generateAcronyms.py
womullan: old habits hard to break

bin/generateAcronyms.py
womullan: good catch total error !

bin/generateAcronyms.py
womullan: ok

bin/generateAcronyms.py
womullan: actually i think this is not needed - there was a ,,,,, in the definition file giving a blank entry in the definitions ..

bin/generateAcronyms.py
womullan: ok mistakenly removed ..

womullan:
ok flake fixed and the one mistake with sub_line

PR #330, opened 2021-01-27
Check correct number of cols

womullan:


mfisherlevine (commented):
The code look fine. Only issues are typos in the comments (`# now strick no blanks and 6 cols`), the commit message, and the PR title.

It might also be worth raising a `ValueError` or something like that, rather than the super-broad `Exception`, but given this is a tex package, it also probably doesn't matter much.

PR #338, opened 2021-02-10
add google api

womullan:


requirements.txt
```diff
@@ -1 +1,5 @@
 pyyaml
+google-api-python-client 
+google-auth-httplib2 i
```
timj, line 3: This extra `i` is a problem.

womullan: i think anyway for now I will not add these - I will do it in the doc first and make sure it works
not all doc suse google api

PR #352, opened 2021-05-04
just catch unescaped #

womullan:


timj (approved):

womullan:
i used raw since it has \ and who knows what else later ..

timj:
Sorry that my comments seem to be detached from the code. I made them on the commit and not the PR.

There is no `\` in that raw string.

PR #355, opened 2021-05-10
CLO

womullan:


jonathansick:
Just a note that the use of this acronym is discouraged. The language that should be used in official documents is "Community Forum" or "Vera C. Rubin Community Forum".

PR #372, opened 2021-07-29
add tocstyle for mac2021 tex

womullan:
I prefer to  use the fix and drop tocstyle but i guess 1 step at a time ..


timj (approved):
Seems like a reasonable compromise for now. It's not hard to update the lsst-texmf deployment but this gets things fixed quickly.

PR #381, opened 2021-10-06
 added gogolers and burwood, make db2authors work for adass

womullan:


bin/db2authors.py
```diff
@@ -47,19 +47,34 @@
...
 args = parser.parse_args()
 
-SPIE = False
+buffer_affil = False
+buffer_authors = False
 affil_cmd = "affiliation"
+affil_form = r"\{}[{}]{{{}}}"
+auth_afil_form = "{}{}{}"
+author_form = r"\author{}{{{}~{}}}"
+author_super = False
```
timj, line 56: Can you add some commentary as to what the different booleans affect?

bin/db2authors.py
```diff
@@ -47,19 +47,34 @@
                          'verbose' displays all the information...""")
 args = parser.parse_args()
 
-SPIE = False
+buffer_affil = False
```
timj, line 50: Add comment that this is the aas setting? It may be worth being explicit against future choices with

```python
if args.mode == "aas":
    ...
elif args.mode == "spie":
    ...
elif args.mode == "adass":
    ...
else:
    raise RuntimeError(f"Unrecognized author mode: {args.mode}")
```

bin/db2authors.py
```diff
@@ -91,41 +106,55 @@
...
 
-    if SPIE:
+    if buffer_affil:
         orcid = "[{}]".format(affilAuth)
     else:
         if "orcid" in auth and auth["orcid"]:
             orcid = "[{}]".format(auth["orcid"])
 
+    orc = ""
+    if "orcid" in auth and auth["orcid"]:
```
timj, line 153: Can write both if-clauses as:
```python
orc = auth.get("orcid", "")
email = auth.get("email", "")
```

bin/db2authors.py
```diff
@@ -135,15 +164,60 @@
...
-    if SPIE:
-        print(*affilOutput, sep="\n")
+    # adass has index and paper authors ..
+    addr = affil[affilset[0]].split(',')
+    tute = addr[0]
+    ind = len(addr) - 1
+    if ind > 0:
+        country = addr[ind]
+        ind = ind - 1
+    if ind > 0:
```
timj, line 174: I can't help but feel that there's some cleverer logic here with the multiple decrements of `ind` but I can't think of anything immediately.

bin/db2authors.py
womullan: but all the aas ones do NOT specify aas .. its is the original default

PR #387, opened 2021-11-05
 fix for  comma on last autho in ADASS format

womullan:


bin/db2authors.py
```diff
@@ -119,6 +119,7 @@
 pAuthorOutput = list()
 indexOutput = list()
 
+anum = 0
 for authorid in authors:
```
timj, line 123: You could use:

```python
for anum, authorid in enumerate(authors):
    ...
```
and then you wouldn't need to count authors yourself.

womullan: that is black magic i never saw before .. so it just counts because ??

womullan: done

PR #390, opened 2021-12-02
Add some docs about working with a submodule.

ktlim:


docs/submodule.rst
```diff
@@ -42,5 +42,58 @@ When making a fresh clone of the repository, you will have to execute this withi
...
+
+Keep in mind that this will also update the texmf code, so you may have to resolve any incompatibilities.
+When you commit any changes to your repo, the new submodule hash will be committed as well.
+
+If you decide you want to revert this update before committing it, you execute:
+
+.. code-block:: bash
+
+    cd lsst-texmf
+    git checkout $(cat ../.git/modules/lsst-texmf/HEAD)
```
timj, line 64: I thought you could do `git checkout lsst-texmf` and that reverted it to the original commit.

docs/submodule.rst
```diff
@@ -42,5 +42,58 @@ When making a fresh clone of the repository, you will have to execute this withi
...
+
+If you would like to edit the lsst-texmf submodule in place, rather than cloning it separately, for example to add a bibliography reference, you execute:
+
+.. code-block:: bash
+
+    cd lsst-texmf
+    git checkout main
+
+Then pull, edit, commit, push, and merge as usual within the submodule.
+Again, when you commit any changes to your repo, the new submodule hash will be committed as well.
```
timj, line 89: Maybe add that things can get very confused if you commit the submodule change in the parent repo but forgot to push the lsst-texmf changes back upstream.

docs/submodule.rst
timj: in fact it turns out that you need `git submodule update` to do that without going into the subdirectory.

PR #391, opened 2021-12-02
Add docs on working with submodules

jdswinbank:
@ktlim beat me to it! (#390)

### DIFF
--- a/docs/submodule.rst
+++ b/docs/submodule.rst
@@ -23,7 +23,11 @@ Typically, you can do this exporting the ``TEXMFHOME`` at the beginning of the M
 
 .. code-block:: make
 
-   export TEXMFHOME = lsst-texmf/texmf
+   export TEXMFHOME ?= lsst-texmf/texmf
+
+.. note::
+
+   Assignment using ``?=`` means you can choose another lsst-texmf tree by setting ``TEXMFHOME`` in your environment before running ``make``.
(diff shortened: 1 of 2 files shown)


docs/submodule.rst
```diff
@@ -23,7 +23,11 @@ Typically, you can do this exporting the ``TEXMFHOME`` at the beginning of the M
 
 .. code-block:: make
 
-   export TEXMFHOME = lsst-texmf/texmf
+   export TEXMFHOME ?= lsst-texmf/texmf
+
+.. note::
+
+   Assignment using ``?=`` means you can choose another lsst-texmf tree by setting ``TEXMFHOME`` in your environment before running ``make``.
```
jonathansick, line 30: The interesting thing about this is that it means documents built in CI never use the lsst-texmf specified in the submodule; but instead in the docker image (which is always the `main` brach of lsst-texmf) because that docker container specifies TEXMFHOME. So it kind of defeats the point of the submodule, right?

jdswinbank: My first instinct is that that sounds like an unfortunate decision decision in the Docker image: it brings into tension the needs of CI and the needs of interactive use on a local system, which doesn't seem great. Luckily, it's no longer my problem! :-)

timj:
Closed by #390

PR #392, opened 2021-12-14
Added Peter Ferguson and Erik Dennihy

leannep:


psferguson (approved):

arunkannawadi (approved):

etc/authordb.yaml
```diff
@@ -1358,6 +1372,13 @@ authors:
     initials: Bryce
     name: Kalmbach
     orcid: 0000-0002-6825-5283
+  kanawadia:
```
arunkannawadi, line 1375: I _think_ the username is correct though. I remember having an unfortunate typo

leannep: I can update this as well

etc/authordb.yaml
```diff
@@ -2077,6 +2084,13 @@ authors:
     initials: Andrew P.
     name: Rasmussen
     orcid: null
+  reeds:
+    affil:
+    - Princeton-Astro
+    altaffil: []
+    initials: Sophie
```
sr525, line 2098: I use Sophie L Reed because Scott Reed publishes too many papers on soil and makes me feel bad.

leannep: OK, will correct

PR #397, opened 2021-12-30
DM-33065: Generate automated bibfile from lsst.io

womullan:


womullan:
I would like to merge the new bibfile now it works
then consider the best action for update. Make lsst.bib will update the file merging the static entries with all the lsst.io ones.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env python3
+
+r"""Utility that interogates ook (lsst.io) to get doc metadata
```
timj, line 3: Doesn't need to be a raw string.  Typo "interogates" -> "interrogates"

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+then generates a bibfile fpr inclusion in texmf.
```
timj, line 4: "fpr" -> "for"

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+from datetime import datetime
+import sys
+import calendar
+
+MAXREC = 5000
+
+
+def write_latex_bibentry(auth, title, year, month, handle, fd=sys.stdout):
+    """ Write a bibentry for document info passed
+    Parameters
```
timj, line 18: Needs blank line above. Probably should also clarify in docstring that this is specifically for documents that have lsst.io entries (since the URL is assuming that).

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+    ----------
+    auth : `str``
+        Author list ready for tex
```
timj, line 21: Does this mean "F. Author1 and B. Author2" format that bibtex likes?

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+    print("    title = {{{}}},".format(title), file=fd)
+    print("     year = {},".format(year), file=fd)
+    print("    month = {},".format(month), file=fd)
+    print("   handle = {{{}}},".format(handle), file=fd)
+    print("      url = {{http://{}.lsst.io }} }}".format(handle), file=fd)
+
+
+def generate_bibfile(outfile, query):
+
+    client = SearchClient.create('0OJETYIVL5', 'b7bd2f1080a5c4fe5eee502462bcc9d3')
```
timj, line 40: What is the significance of these strings? Are they public information?

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+        authors = " and "
+        authors = authors.join(d['authorNames'])
+        dt = d['sourceUpdateTimestamp']
+        date = datetime.fromtimestamp(dt)
+        month = calendar.month_abbr[date.month].lower()
+        write_latex_bibentry(checkFixAuthAndComma(fixTexSS(authors)), fixTex(d['h1']),
+                             date.year, month, d['handle'], outfile)
+        print(file=outfile)
+
+    print(f"Got {count} records max:{MAXREC} produced {bcount} bibentries to {outfile}")
```
timj, line 76: I was assuming that MAXREC was maximum results per page and that there must be an API to iterate over paged results? I assume it's not needed because we don't have 5000 documents.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+    print("@DocuShare{{{},".format(handle), file=fd)
+    print("   author = {{{}}},".format(auth), file=fd)
+    print("    title = {{{}}},".format(title), file=fd)
+    print("     year = {},".format(year), file=fd)
+    print("    month = {},".format(month), file=fd)
+    print("   handle = {{{}}},".format(handle), file=fd)
+    print("      url = {{http://{}.lsst.io }} }}".format(handle), file=fd)
+
+
+def generate_bibfile(outfile, query):
```
timj, line 38: Maybe document what the query string looks like?

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+          '## Add static entries in etc/static_entries.bib (or remove them if they clash.\n'
+          '## This files should contain ALL entries on www.lsst.io ', file=outfile)
+
+    bcount = 0
+    for count, d in enumerate(res['hits']):
+        if d['series'] == "TESTN":
+            continue
+        bcount = bcount + 1
+        authors = " and "
+        authors = authors.join(d['authorNames'])
```
timj, line 68: I don't think the previous line is helping clarity because it's not initially set to authors at all but " and ".
Maybe just write:

```python
authors = " and ".join(d["authorNames"])
```
?

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+
+    res = index.search(query, params)
+
+    print('## DO NOT EDIT THIS FILE. It is generated from generateBibfile.py\n'
+          '## Add static entries in etc/static_entries.bib (or remove them if they clash.\n'
+          '## This files should contain ALL entries on www.lsst.io ', file=outfile)
+
+    bcount = 0
+    for count, d in enumerate(res['hits']):
+        if d['series'] == "TESTN":
```
timj, line 64: A general comment that the coding standard says we should try to be consistent in single vs double quotes for strings in a single file. Not important for now but consider in the future.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+        date = datetime.fromtimestamp(dt)
+        month = calendar.month_abbr[date.month].lower()
+        write_latex_bibentry(checkFixAuthAndComma(fixTexSS(authors)), fixTex(d['h1']),
+                             date.year, month, d['handle'], outfile)
+        print(file=outfile)
+
+    print(f"Got {count} records max:{MAXREC} produced {bcount} bibentries to {outfile}")
+
+
+def fixTex(text):
```
timj, line 79: docstring to say what this is trying to fix would be good. It's fixing the ASCII characters that need to be escaped from tex.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+    try:
+        text.encode('ascii')
+    except UnicodeEncodeError:
+        for ci, co in [('’', "'"),
+                       ('…', '...'),
+                       ('…', '...'),
+                       ('“', '"'),
+                       ('”', '"'),
+                       ('´', "'"),
+                       (' ', ' '),
```
timj, line 102: I'm not sure what this line is doing because when I copy and paste the first value it seems to copy as a simple space. Is it meant to be some special unicode space? (in which case it might be better to define it explicitly in unicode.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+    return text
+
+
+def checkFixAuthAndComma(authors):
+    if "," in authors and "and" not in authors:
+        # a bit heav handed but
+        authors = authors.replace(",", "and")
+
+
+def fixTexSS(text):
```
timj, line 92: Add docstring to explain what this is for.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+    try:
+        text.encode('ascii')
+    except UnicodeEncodeError:
+        for ci, co in [('’', "'"),
+                       ('…', '...'),
+                       ('…', '...'),
```
timj, line 98: Are these two different ellipsis characters? The line above looks the same and if I copy and paste they are identical.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+
+def checkFixAuthAndComma(authors):
+    if "," in authors and "and" not in authors:
+        # a bit heav handed but
+        authors = authors.replace(",", "and")
+
+
+def fixTexSS(text):
+    try:
+        text.encode('ascii')
```
timj, line 94: Maybe comment that this is here to short circuit the replacement since if the encoding works then there aren't any non-ascii characters. It would also serve to comment that the encoded bytes are irrelevant.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+                       ('í', "\\'{i}"),
+                       ('ó', "\\'{o}"),
+                       ('ñ', '\\~{n}'),
+                       ('ö', '\\"{o}'),
+                       ('û', '\\r{u}'),
+                       ('ü', '\\"{u}'),
+                       ('ù', "\\`{u}"),
+                       ('ž', '{\\v z}'),
+                       ('Ž', '{\\v Z}'),
+                       ('􏰎', ' '),
```
timj, line 126: This line looks odd. It comes out as a space here but copies as a ? in a square (which is above already).

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+                       ('ï', '\\"{i}'),  # really i dieresis
+                       ('ô', '\\r{o}'),
+                       ('–', '-'),
```
timj, line 129: This is an en-dash (which is fine) but I don't see em-dash anywhere.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+                       ('‘', "'"),
+                       ('ʻ', '\''),
```
timj, line 131: These two lines are different but don't they translate to the same thing -- maybe best to use double quotes as is done above.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def fixTexSS(text):
+    try:
+        text.encode('ascii')
+    except UnicodeEncodeError:
+        for ci, co in [('’', "'"),
+                       ('…', '...'),
+                       ('…', '...'),
+                       ('“', '"'),
```
timj, line 99: Maybe add a comment at the end of this line and below to note that these are in fact two distinct double quote unicode characters. 8221 and 8220 (LEFT and RIGHT) even though in my browser window they seem to be the same.

etc/static_entries.bib
```diff
@@ -0,0 +1,1703 @@
+## Add or remove entries from etc/static_entries.bib for docushare items not appearing in lsst.io
```
timj, line 1: In an ideal world this would be a standalone bib file (docushare.bib)  but I guess it would be too annoying to change all the documents to read an extra bib file.

timj (approved):
Looks good. Some minor comments.

bin/generateBibfile.py
womullan: yes and another place i replace the commas .. but it depends a bit on what people put in the documents .. I am not trying to swizle their inputs

bin/generateBibfile.py
womullan: good point @jonathansick  do I need to put these in a secret ?

bin/generateBibfile.py
womullan: yes we have ~500 recs so I can ignore paging and get them in one pop

bin/generateBibfile.py
womullan: '􏰎 inherited from robert like that ...I did not play with it

bin/generateBibfile.py
womullan: again inherited I agree I will remove one

bin/generateBibfile.py
womullan: I am sure RHL did more than I need I could probably remove some of these - on the other hand he did have to deal with them apparently. I added a comment on this

bin/generateBibfile.py
womullan: added

etc/static_entries.bib
womullan: that was my reasoning ...

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+then generates a bibfile fpr inclusion in texmf.
+"""
+
+import argparse
+from algoliasearch.search_client import SearchClient
+from datetime import datetime
+import sys
+import calendar
+
+MAXREC = 5000
```
jonathansick: I think by default this is limited to 1000 https://www.algolia.com/doc/api-reference/api-methods/search/#about-this-method

bin/generateBibfile.py
jonathansick, line 75: I'd suggest also testing if series is in the hit result. At some point we'll start adding records that don't have handles and those will need to treated differently:

```suggestion
        if 'series' in d.keys() or d['series'] == 'TESTN':
```

For extra robustness, you might also want this script to opt into specific series rather than only excluding specific series. i.e. `d['series'] not in allowed_series` where `allowed_series` is a list of known document handles.

bin/generateBibfile.py
jonathansick: From a code maintenance perspective, I'd also suggest a re-write here. Preping the authors list here, and then running `checkFixAuthAndCommand` later means there's two places the author text is being prepared. I think one function to format an author text is a good idea.

bin/generateBibfile.py
jonathansick: Having both `fixTex` and `FixTexSS`, and the fact that they don't relate to each other seems like a potential issue.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+        if d['series'] == "TESTN":
+            continue
+        bcount = bcount + 1
+        authors = " and "
+        authors = authors.join(d['authorNames'])
+        dt = d['sourceUpdateTimestamp']
+        date = datetime.fromtimestamp(dt)
+        month = calendar.month_abbr[date.month].lower()
+        write_latex_bibentry(checkFixAuthAndComma(fixTexSS(authors)), fixTex(d['h1']),
+                             date.year, month, d['handle'], outfile)
```
jonathansick, line 83: For reliability, I'd recommend doing a check on any attributes and ensuring they're present. I'm not 100% sure that Algolia guarantees that by setting attributesToRetrieve that all records will have these attributes. I tried to find docs on this, but I couldn't find an answer.

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+import argparse
+from algoliasearch.search_client import SearchClient
+from datetime import datetime
+import sys
+import calendar
+
+MAXREC = 5000
+
+
+def write_latex_bibentry(auth, title, year, month, handle, fd=sys.stdout):
```
jonathansick, line 16: A good approach here would be to create a class instance for each record. That class could have methods for preparing the algolia data into the right format for bibtex and also have a method for outputting the result. That's a great way to avoid having functions like this with a lot of arguments.

jonathansick (commented):
I've included some comments that I hope will improve the reliability of this script. While some comments are judgement calls, the ones prefixed with ⚠️ are bugs that I think need to be fixed.

Overall, I do have some concerns about the reliability and maintainability of the current algorithm so it would be good to both have the script work more defensively for any irregularities in the data from Algolia.

One thing that would be a huge improvement is to actually parse and write out the bib from Python objects using [pybtex](https://docs.pybtex.org/api/index.html) rather than writing straight to text and concatenating with `cat`. In particular, I think having a normalized sort order for the bibfile will be useful in identifying any regressions in the data using git commit diffs.

bin/generateBibfile.py
jonathansick: They're public information, though it'd be good to make them variables so it's easier to update them in the future.

.github/workflows/generatebib.yaml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -r requirements.txt
+          python -m pip install ltd-conveyor
+
+      - name: Pull existing images
+        run: |
+          docker pull lsstsqre/lsst-texmf:base || true
+          docker pull lsstsqre/lsst-texmf:dependencies || true
+          docker pull lsstsqre/lsst-texmf:latest || true
```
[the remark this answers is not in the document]
womullan: you mean no docker pulls ok

.github/workflows/generatebib.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Generate bibfile
+        run: |
+          docker run --rm -v `pwd`:/workspace -w /lsst-texmf lsstsqre/lsst-texmf:latest sh -c 'make lsst.bib'
+
+      - name: push changes to github
```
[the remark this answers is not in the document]
womullan: last version is a PR (intends to be)

Makefile
```diff
@@ -59,6 +59,11 @@ docs: $(PDF) $(TESTS)
 	cp etc/glossarydefs.csv docs
 	make -C docs html
 
+lsst.bib: 
+	bin/generateBibfile.py texmf/bibtex/bib/lsst.bib
+	cat etc/static_entries.bib >> texmf/bibtex/bib/lsst.bib
```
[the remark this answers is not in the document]
womullan: yes there may be dup keys - the PR will then fail.
 Though this should be very few only the few reamaining lsst.io entries as they show up in ook. 
I did not want to sort across the two files. I did mean to ask you if you can have a sort order on the query - I guess you are indicating the order returned from ook is random ..

bin/generateBibfile.py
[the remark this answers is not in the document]
womullan: added

bin/generateBibfile.py
```diff
@@ -0,0 +1,162 @@
...
+
+    """
+
+    print("@DocuShare{{{},".format(handle), file=fd)
+    print("   author = {{{}}},".format(auth), file=fd)
+    print("    title = {{{}}},".format(title), file=fd)
+    print("     year = {},".format(year), file=fd)
+    print("    month = {},".format(month), file=fd)
+    print("   handle = {{{}}},".format(handle), file=fd)
+    print("      url = {{http://{}.lsst.io }} }}".format(handle), file=fd)
```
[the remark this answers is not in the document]
womullan: ah yes I had this routine from elsewhere sans url but of course lsst.io gives a url nice !

bin/generateBibfile.py
womullan: i want them all only TESTN  do not form good entries ..  i have taken that bit with an AND nor an OR .. the or would still check the second term even if the first failed.

bin/generateBibfile.py
womullan: agreed - it is why i asked if you had an ook client library .. with such classes defined.

bin/generateBibfile.py
womullan: If not present they should come out as None .. which is probably ok. But i will see about adding a check/report

bin/generateBibfile.py
womullan: do variables for the ook application id and index already exist ? it seems like they should be org level not repo level

womullan:
I am going to merge this now .. and may work on it some more - I need to merge for the action to show up

PR #486, opened 2022-04-21
Added Eli Rykoff

leannep:


etc/authordb.yaml
```diff
@@ -2615,6 +2615,13 @@ authors:
     initials: Stefano
     name: Russo
     orcid: null
+  erykoff:
```
erykoff, line 2618: I think this should be `rykoffe` to go with the standard.

leannep: fixed

erykoff (approved):

PR #500, opened 2022-05-04
U/womullan/pdm

womullan:


etc/authordb.yaml
```diff
@@ -2377,7 +2377,7 @@ authors:
     affil:
     - CCA
     altaffil: []
-    initials: adrian
+    initials: Adrian
     name: Price-Whelan
     orcid: null
```
timj, line 2382: Looks like his ORCID is 0000-0003-0872-7098  (https://orcid.org/0000-0003-0872-7098)

womullan: will catch in the next round

PR #505, opened 2022-05-09
add static bib for LSE-479

jhoblitt:


timj (approved):

etc/static_entries.bib
```diff
@@ -1279,6 +1279,14 @@ @DocuShare{LSE-400
   url = {https://lse-400.lsst.io},
 }
 
+@DocuShare{LSE-479,
+  author =       {Network Engineering Team (NET)},
```
timj, line 1283: Actually, you might need a couple of braces around the author to stop it being "N. E. Team" in latex.

jhoblitt: It looks like many of the other declarations are assuming that it is a verbatim string (no idea if that is a correct assumption). E.g.
```
  author =       {Tim Jenness and John Swinbank and Simon Krughoff and
                  Gregory Dubois-Felsmann and David Ciardi},
```

timj: No. "Tim Jenness and John Swinbank" will be parsed by the bibtex system and depending on the BST file it could come out as "Jenness, T., & Swinbank, J" or as anything else. "and" is special and it assumes "first last". If you say "Network Engineering Team" it will likely appear in the file as "Team, N. E.". You need to say "{Network Engineering Team}" to make it print that in the references.

jhoblitt: There are a couple of examples of the author being quoted so I've added nested `{}`s.

PR #507, opened 2022-05-10
add workflow_dispatch to dockercache triggers

jhoblitt:
'cause some people are impatient...

### DIFF
--- a/.github/workflows/dockercache.yaml
+++ b/.github/workflows/dockercache.yaml
@@ -6,6 +6,7 @@
 name: Refresh Docker cache
 
 on:
+  workflow_dispatch:
   schedule:
     - cron: "0 4 * * MON"
 


jhoblitt:
There we go... docker hub is now showing the last push to master from `ci.yaml`.  This may not be completely necessary but still seems like a nice-to-have.

timj:
I have no idea what "workflow_dispatch" does so not entirely sure I can review.

womullan:
I had ont appreciated that docker was caching the base image - so this forces it to refhresh whereas normal CI would just use the exisiting base ?  I guess that is good ...  otherwise how was this ever getting updated -- Jonathan ocasionally did it ?

womullan (approved):
this should also be documented (in dveloper.lsst.io?).. so we know what it is later

timj:
I think @jonathansick set all this up so he probably should comment.

jonathansick (dismissed):
I'm in principle ok with adding a `workflow_dispatch` to refresh the base image on-demand. However, your PR isn't much to go on — what is making you need to refresh the base image? I forsaw a weekly schedule as being sufficient and don't really see why we'd need folks to have a more active enagement. Could you explain @jhoblitt ?

womullan:
> I'm in principle ok with adding a `workflow_dispatch` to refresh the base image on-demand. However, your PR isn't much to go on — what is making you need to refresh the base image? I forsaw a weekly schedule as being sufficient and don't really see why we'd need folks to have a more active enagement. Could you explain @jhoblitt ?

I believe @jhoblitt was faced with a problem in the base image causing an issue in the CI and update was needed to fix it. For some repos heavily dependent on the image for testing the ability to force an update may be occasionally useful. So I would move to rebase and merge this.

jonathansick:
I'm closing as I don't think this is needed and there hasn't been any additional discussion since the initial issues were raised.

jonathansick:
@womullan ok, if you think it'll help that's ok then.

jhoblitt:
I was away yesterday.  I concur with @wil that have the ability to manually trigger a run is desirable, even if it isn't strictly necessary.

PR #520, opened 2022-05-19
DM-34607: Add note field back to DocuShare bibtex entries

timj:
These used to be present until the code was switched to
auto-generation. Arguably this should be "howpublished"
and "publisher" but the Sphinx bibtex plugin for now assumes
that there is a "note" field.

This makes it clear what the document is and where it comes from.


timj:
I'm not really sure if the CI runs the bibtex file generation code.

PR #546, opened 2022-06-15
more participants in DSS workshop

rmandelb:
This PR adds a few more participants from the Data to Software to Science workshop to the author DB.

To include a participant with Serbian accents in his name (new entry `savicd` in the database), I had to use some non-default latex accents, which are available when you use `\usepackage[T1]{fontenc}` in your latex document.  Is that acceptable in general or do you prefer to use only default latex accents?  If it is acceptable, how should this be documented?

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -3,7 +3,7 @@ affiliations:
   BNL: Brookhaven National Laboratory, Upton, NY 11973, USA
   Belgrade-AO: Astronomical Observatory, Volgina 7, P.O.\ Box 74, 11060 Belgrade,
     Serbia
-  Belgrade-Uni: University of Belgrade-Faculty of Mathematics, Department of astronomy,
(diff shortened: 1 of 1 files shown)


womullan (approved):
unless there is a way to get those chars without the package i think it is ok and it only for papers with authors with those accents.

we can document directly in docs of this package  https://lsst-texmf.lsst.io/developer.html#contributing-documentation  
though i note the authors db does not have a section ! @timj  may have it elsewhere

A comment directly near savicd in the authordb would also be good.

timj:
The answer depends a bit on the journal being used. If these are lsstdoc documents we can put any default font directives directly into our class. If it's AAS then it may depend on what their rules are. In theory modern latex supports unicode so you can try using unicode characters directly in the yaml. If AAS let you put extra font encoding directives at the top of your latex file then that's fine with me.

rmandelb:
Merging #546 after addressing feedback (comment added in author DB).

PR #550, opened 2022-06-17
U/rmandelb/authors

rmandelb:
This PR has some updates to the authors DB:
- Correcting how accents show up for two people (my original incantation led to misplaced accents in two cases).
- Adding some participants from the DSS workshop, and fixing/updating the entries for others at their request.


womullan (approved):
thank you for doing these !

PR #551, opened 2022-06-17
U/rmandelb/authors

rmandelb:
Sorry @womullan - I had finished all the input to the author form, so I made the previous PR.  But then I found one more (not requested via the form... somebody directly edited the authors.tex again).


womullan (approved):
for small corrections like this an approve is not needed - if the build passes you may mere

PR #560, opened 2022-06-26
U/rmandelb/authors

rmandelb:
This PR adds some additional authors for the DSS white paper (via the form and capturing direct edits to authors.tex as well).  Probably it's a big enough PR to be worth a quick review - I'm not sure I'd call it a bugfix.


womullan (approved):
looks good - regeenrated the authors in the paper and it also looks ok.

PR #566, opened 2022-07-02
OO bib entry 

womullan:
not ready yet

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,12 +18,18 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install flake8 pybtex
+          python -m pip install -r requirements.txt
 
       - name: Run flake8
         run: flake8
 
       - name: Run pybtex
         run: make test-pybtex
+
+      - name: Run pytest
+        env:
+          PYTHONPATH: bin
+        run: pytest
 
   build:
     runs-on: ubuntu-latest
--- a/bin/bibtools.py
+++ b/bin/bibtools.py
@@ -25,7 +25,7 @@ class BibEntry:
     # not all entries have all fields
     """
 
-    def __init__(self, author, title, month, handle, year, note="", url=""):
+    def __init__(self, author=None, title=None, month=None, handle=None, year=None, note="", url=""):
         self.author = author
         self.title = title
         self.month = month
@@ -34,7 +34,7 @@ class BibEntry:
         self.url = url
         self.year = year
         self.type = "@DocuShare"
-        if "" == note:
+        if "" == note and handle:
             prefix, _ = self.handle.split("-", 1)
             series = TN_SERIES.get(prefix, "")
             self.note = f"{{Vera C. Rubin Observatory {series}}}"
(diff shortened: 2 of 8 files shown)


womullan:
@jonathansick now this is used in the generatebib.  For a first step I think its good. Pybtex runs over the entire thing but I guess in another improvement i could parse the static file and sort the entire bib instead of just concatinating - though concat also keeps clean the generated from the static..

womullan:
@jonathansick ok for this ??   dont want to get too out of date with main

bin/bibtools.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
+
+"""This class represents a bibtex bibentry.
+it can construct from a string and output to the bib file as needed.
+it is comparable hence sortable.
```
jonathansick, line 5: I don't quite follow how `BibEntry` comparisons work? Wouldn't it need `__eq__`, `__le__`, etc [rich comparison](https://docs.python.org/3/reference/datamodel.html#object.__lt__) methods?

bin/bibtools.py
```diff
@@ -0,0 +1,57 @@
...
+    "SMTN": "Simulations Team Technical Note",
+    "SQR": "SQuaRE Technical Note",
+    "DMTR": "Data Management Test Report",
+    "LDM": "Data Management Controlled Document",
+}
+
+
+class BibEntry:
+    """A class to represent a BibEntry
+    # not all entries have all fields
```
jonathansick, line 25: Maybe a typo in the docstring with the line comment syntax?

bin/bibtools.py
```diff
@@ -0,0 +1,57 @@
...
+        self.year = year
+        self.type = "@DocuShare"
+        if "" == note and handle:
+            prefix, _ = self.handle.split("-", 1)
+            series = TN_SERIES.get(prefix, "")
+            self.note = f"{{Vera C. Rubin Observatory {series}}}"
+
+    def write_latex_bibentry(self, fd=sys.stdout):
+        """ Write a bibentry for document info passed.
+        This assumes its an lsst.io entry and constructs the url accordingly.
```
jonathansick, line 44: Is this comment about URLs from an earlier version of the code? `BibEntry` doesn't seem to be doing any URL formatting.

.github/workflows/ci.yaml
```diff
@@ -18,13 +18,19 @@ jobs:
...
 
       - name: Run flake8
         run: flake8
 
       - name: Run pybtex
         run: make test-pybtex
 
+      - name: Run pytest
+        env:
+          PYTHONPATH: bin
```
jonathansick, line 31: I see this as a bit of a code smell for packaging issues, though it's not worth changing now. If we get a chance we can come back and put in a more standard infrastructure around lsst-texmf's Python tools.

jonathansick (approved):
Seems ok.

bin/bibtools.py
womullan: yes i have not done that yet .. future effort

womullan: comparators added

PR #570, opened 2022-07-07
U/rmandelb/authors

rmandelb:
More authors, affils in authordb.yaml.  This may be the last batch for the white paper, though a few more could trickle in later.


womullan (approved):
looks good

PR #585, opened 2022-07-29
one more affil, two more co-authors

rmandelb:
A tiny PR for a few more authors in the authorDB.


womullan (approved):
Looks good - I made a few glossary tweeks as well I may be done

PR #648, opened 2022-10-27
add python3-pygments package

jhoblitt:
The `pygmentize` command is required by the latex minted package.


jonathansick (approved):
It might be preferable to install pygments via `requirements.txt` instead of through the Ubuntu package manager, but either way is fine.

womullan:
yes its nice for local execution if requirements.txt has it - but both can be done

womullan (approved):

jhoblitt:
Rebasing on current master.

jhoblitt:
@jonathansick I don't have strong feelings about which mechanism is used. I typically have a slight preference for OS provided packages.

PR #789, opened 2023-05-19
Tickets/dm 35537 - fix for null altaffil in authors (plus test now added and run)

womullan:


bin/db2authors.py
```diff
@@ -228,7 +228,7 @@ def get_initials(initials):
         if buffer_affil:
             print(*affilOutput, sep="\n")
         else:
-            if "altaffil" in auth:
+            if "altaffil" in auth and auth["altaffil"] is not None:
```
timj, line 231:
```suggestion
            if auth.get("altaffil"):
```

bin/validate_authors.py
```diff
@@ -0,0 +1,29 @@
...
+"""
+
+import os
+import yaml
+import subprocess
+
+
+def make_all(authordb):
+    """Go through all authors and put in one big file so we can test"""
+    with open("authors.yaml", "w") as file:
```
timj, line 12: I don't understand the purpose of writing this file. What's it used for?

bin/validate_authors.py
```diff
@@ -0,0 +1,29 @@
...
+        for a in authordb["authors"]:
+            print(f" - {a}", file=file)
```
timj, line 14: Normally you would create a python dict or list and then use `yaml.dump` to ensure you are writing compliant yaml. Writing the yaml yourself means you have to worry about possible quoting issues in the author names.

bin/validate_authors.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def main():
+    dbfile = os.path.normpath(
+        os.path.join(".", "etc", "authordb.yaml"))
```
timj, line 19: Strange line break here.

bin/validate_authors.py
womullan: to call db2auth for all authors thus checking there are no errors like the None in altaffil - you only get that if that oathor is included

bin/validate_authors.py
womullan: ok but its a single value loaded from a yaml file which is the key in that file .. if you insist i will add a List instead ..

womullan: list used now

PR #813, opened 2023-06-16
Update lsst.bib

fedhere:
added brackets in the author of PSTN-055 to render correctly (otherwise the citation renders as Committe 2023)


womullan (changes requested):
this will not work .. its a generted file - read the docs ..
Open a ticket and assign to me I will fix the plumbing

PR #942, opened 2023-11-13
 A few new SLAC acronyms

womullan:


etc/glossarydefs.csv
```diff
@@ -220,6 +222,7 @@ DAA,Dual Axis Actuator,TS,,,A
 DAC,Data Access Center,LSST DM,,,A
 DAF,data access framework,TS,Document-11936,,A
 DALI,Data Access Layer Interface (IVOA standard),VO,,,A
+DAPC,Departmental Appointments and Promotion Committee,OPS,,,A
```
drphilmarshall, line 225:
```suggestion
DAPC,SLAC FPD Departmental Appointments and Promotion Committee,OPS,,,A

```

drphilmarshall: I think this only exists in the SLAC FPD context so maybe include those here? OK to leave them out too I think, up to you.

drphilmarshall (approved):
Thanks Wil! Youcould add "SLAC FPD" before "DAPC" just to make teh context clear, perhaps. I left a suggestion for you to take or leave.

plazas:
Should the Spanish version be updated as well when new terms are added, so it doesn’t go out too out of sync?

womullan:
I would not be sure how to translate dome of these office names .. but in principle yes we should try to keep in synch --- I tend to add a couple a week ..

womullan:
@plazas I could  add a function to the script to output terms with no translation

PR #945, opened 2023-11-17
 CRAC CRAH

womullan:


plazas (approved):
Looks good, thanks!  I would say "computadores" instead of "ordenadores" (both are correct though).

PR #1007, opened 2024-01-30
Update authordb.yaml - Adding Bekah Polen

gmegh:


womullan (changes requested):
you will have to fix the trainling spaces
also looks like you forked not branched?   You should be able to branch now

womullan:
i have added this in https://github.com/lsst/lsst-texmf/pull/1017

PR #1016, opened 2024-02-06
Add myself to authordb.yaml

bjwhite-fnal:


etc/authordb.yaml
```diff
@@ -638,6 +638,13 @@ authors:
     initials: Rahul
     name: Biswas
     orcid: null
+  bjwhite:
```
womullan, line 641: convention is LastnameInitials  so whitebj please

womullan (changes requested):

PR #1047, opened 2024-03-04
Added missing header to fix parse error

jeipollack:
- Added missing header to fix Pandas parse error when reading glossarydefs_es.csv

### DIFF
--- a/etc/glossarydefs_es.csv
+++ b/etc/glossarydefs_es.csv
@@ -42,7 +42,7 @@ AISES,Sociedad Indígeno-Estadounidense de Ciencia e Ingeniería
 AIT,"Ensamblaje, integración y prueba"
 AI&T,"Ensamblaje, integración y prueba"
 AIV,"Ensamblaje, integración y verificación"
-AJ,"The Astronomical Journal" [La Revista de Astronomía]
+AJ,"The Astronomical Journal [La Revista de Astronomía]"
 ALeRCE,Aprendizaje automático para la clasificación rápida de eventos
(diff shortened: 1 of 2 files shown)


etc/glossarydefs_es.csv
```diff
@@ -1,4 +1,4 @@
-English,Español
+English,Español,Subsystem Tags
```
timj, line 1: Which line has the problem? It looks like there is an error in line 45 that GitHub is telling me is the reason it can't read it (the double quoting is confusing it).

womullan: Good to fix the header. That indeed looks like it needs an update - could also drop the " altogether.

womullan (approved):

etc/glossarydefs_es.csv
womullan: I fixed the quote

timj: I don't think the header is wrong. There are only two columns in this file. The third column is caused by the error in row 45. There are no subsystem tags in this file are there?

womullan: some entries have the 3rd column - i had to add it for any entry which is overloaded otherwise i dont know which translation goes wit which entry  see e.g. CAM

timj: Oh yes. Would be clearer if every line had a trailing comma I guess. 😄

timj:
The check is failing because we require rebases rather than merges from main to sync the branch.

PR #1050, opened 2024-03-06
Correct institutional name for IPAC

gpdf:
The formal name we use is "Caltech/IPAC".


timj (approved):
This is fine for your purposes. Will that form be okay for ApJ papers?

gpdf:
@timj 
> [...] Will that form be okay for ApJ papers?

Do you have a recent example of a paper with affiliations generated from that file?  I don't have a clear mental picture of what it looks like.

timj:
The ApJ overview paper for example: https://iopscience.iop.org/article/10.3847/1538-4357/ab042c

It takes that affiliation text and puts it directly in the paper without any attempt to reformat or extract semantics. That was the reason the file was created in the first place.

womullan (approved):

gpdf:
It may look a bit awkward, but if a paper comes out before @jonathansick gets to the refactoring, it'll be OK.

PR #1056, opened 2024-03-09
1055 change the name of my institution lal ijclab and add me in author list

sylvielsstfr:
I have kept the old name LAL for those author still affiliated to LAL in the list.

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -194,6 +194,8 @@ affiliations:
     et L\'{e}onie Duquet, Paris Cedex 13, France
   Paris-Diderot: Universit\'{e} Paris Diderot, Sorbonne Paris Cit\'{e}, F-75013 Paris,
     France
+  Paris-IJCLAB: Laboratoire de Physique des 2 infinis Irène Joliot-Curie – IJCLab - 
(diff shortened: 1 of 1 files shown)


timj:
To clarify, the institution did not change formally from LAL to IJCLAB for everyone, but LAL still exists as an entity that people are still affiliated to?

sylvielsstfr:
Normaly every ex LAL people are in IJCLab with a few exceptions for those that changed to another institute. I have not checked for them. I will go through the authorlist to check this and correct it.

sylvielsstfr:
I have run yamlint on the file and remove the trailing blank.

etc/authordb.yaml
```diff
@@ -639,14 +639,6 @@ authors:
     initials: Rahul
     name: Biswas
     orcid: null
-  blancg:
```
timj, line 642: Can you comment as to why this author was removed from the file?

timj:
I rebased your branch to remove the merge from main.

etc/authordb.yaml
sylvielsstfr: Yes he left LSST

PR #1074, opened 2024-03-25
 urls in static entries also

womullan:


timj:
My main comment is that we add these automatically in our documents because of the special `@DocuShare` handler. Since no-one else has this if they use the exact entries then bibtex falls back to `@Misc`.

When we write the ADS integration we will be sending them entries that work properly for `@Misc` and so include also publisher.

timj (approved):
We can see how it goes. The fact that the publisher and note both have the publisher in is possibly going to lead to some duplication but we shall see how bad it is.

PR #1081, opened 2024-04-02
Update action versions

timj:


timj:
For the record, updating to python 3.11 unblocked documenteer 1.x and sphinx 7.x and those do not work yet with this package. I had to pin to documenteer < 1 and sphinx < 7 before it would work with 3.11. @jonathansick do you know if this is a fundamental problem or something that can be handled with a relatively straightforward change to `docs/conf.py`?

womullan (approved):
good to get these updated

jonathansick:
> @jonathansick do you know if this is a fundamental problem or something that can be handled with a relatively straightforward change to docs/conf.py?

Likely the theme isn't compatible with modern Sphinx. The way to deal with that is to install `documenteer[guide]` and set up `conf.py` as outlined in https://documenteer.lsst.io/guides/configuration.html .

PR #1082, opened 2024-04-03
Automatic bibfile  update

womullan:


timj:
@womullan not sure how we deal with the case where the search results actually missed an existing document and we end up deleting it. I imagine it will come back tomorrow when the search runs again.

womullan:
Yes it will reappear once it is in lsst.io -  the job reproduces lsst.io exactly each night to catch any updates to individual entries.

PR #1091, opened 2024-04-09
Some fixes to bib entries

timj:
We really need to write a script that takes a list of bibcodes and generates these local files. The SPIE entries from a decade ago all had bad page numbers so I had to recopy from ADS.


womullan:
agreed - actually I think there is already a library for doing it somewhere ..

womullan (approved):

PR #1092, opened 2024-04-09
U/womullan/escpercent  Full glossary test - fix double escaping - remove escapes from definiitons

womullan:


bin/generateAcronyms.py
```diff
@@ -371,7 +371,7 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
     }
 
     # findall matches non-overlapping left to right in the order that
-    # we give alternate strings. Therefore when we build the regex
+    # we give alternate strings. Therefor when we build the regex
```
timj, line 374: It was right before...

bin/generateAcronyms.py
```diff
@@ -417,11 +417,13 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
 
 
 def escape_for_tex(str):
-    """Escape tex nasties and return new string"""
-    nstr = str.replace("&", r"\&")
-    nstr = nstr.replace("_", r"\_")
-    nstr = nstr.replace("%", r"\%")
-    return nstr
+    """Escape tex nasties and return new string.
+    But watch out for double whamies .."""
```
timj, line 421:
```suggestion
    But watch out for double whammies."""
```

bin/generateAcronyms.py
```diff
@@ -417,11 +417,13 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
+    text = str
+    for c in specialChars:
+        text = text.replace(c, f"\\{c}")
```
timj, line 424: You can use `rf` so you don't need to double escap.

```suggestion
        text = text.replace(c, fr"\{c}")
```

bin/generateAcronyms.py
```diff
@@ -417,11 +417,13 @@ def find_matches_combo(filename, acronyms, ignore_str=" %"):
...
+        text = text.replace(f"\\\\{c}", f"\\{c}")
```
timj, line 425:
```suggestion
        text = text.replace(fr"\\{c}", fr"\{c}")
```

bin/generateAcronyms.py
```diff
@@ -883,9 +893,16 @@ def dump_gls(filename, out_file):
         help=""" Generate glossary dump file using passed
                                  filename containing  all entries.""",
     )
+    parser.add_argument(
+        "-s",
+        "--skipnone",
+        action="store_true",
+        help=""" Do not load skip acronyms file""",
```
timj, line 900:
```suggestion
        help="""Do not load skip acronyms file""",
```
"skipnone" meaning "skip the skip acronyms file" is a bit tortured. How about "--skip-acronyms/--no-skip-acronyms" defaulting to true and have the variable be "skip_acronyms". That would then allow the code to say "if skip_acronyms" which would be easier to understand.

timj (commented):
Couple of minor comments. My main comment is that `--skipnone` seems non-obvious to me and I suggest an alternative.

timj (approved):

womullan:
skipnone is mainly for testing .. how about skipnoacronyms?

timj:
> "skipnoacronyms"

Ok. That is a bit more explicit than "skipnone". Is this effectively "write all acronyms"?

womullan:
its  ignore skipacronyms file - so yes its writeallacronyms

PR #1114, opened 2024-04-22
Use current guidance for NOIRLab observatory affiliations

timj:


timj:
@ranpalgill / @womullan my understanding from the library page and the newsletter is that Bob and Cathy are now RubinOps and doesn't need a separate NOIRLab affiliation. @ranpalgill I think if you are construction and ops equally then it likely should be RubinOpsC and RubinObsC?

etc/authordb.yaml
```diff
@@ -166,7 +165,8 @@ affiliations:
   NCNRP: National Centre of Nuclear Research, Andrzeja So\l{}tana 7, 05-400 Otwock, Poland
   NCSA: NCSA, University of Illinois at Urbana-Champaign, 1205 W.\ Clark St., Urbana,
     IL 61801, USA
-  NOIRLab: NSF's National Optical-Infrared Astronomy Research Laboratory, 950 N.\ Cherry Ave., Tucson, AZ 85719, USA
+  NOIRLab: NSF NOIRLab, 950 N.\ Cherry Ave., Tucson, AZ 85719, USA
+  NOIRLabC: NSF NOIRLab, Casilla 603, La Serena, Chile
```
timj, line 169: "Casilla 603" does seem to be the contact address listed on the website. Google maps can't find it though so it seems like an odd choice but that is what we are told to use.

ranpalgill:
> @ranpalgill / @womullan my understanding from the library page and the newsletter is that Bob and Cathy are now RubinOps and doesn't need a separate NOIRLab affiliation. @ranpalgill I think if you are construction and ops equally then it likely should be RubinOpsC and RubinObsC?

I am officially RubinOpsC

womullan (approved):
thanks looks good

PR #1119, opened 2024-04-25
Automatic bibfile  update

womullan:


timj:
What a strange PR this was. It moved lots of months forward by one as if there is a timezone involved somewhere... Ah. I think it might be that the timestamp calculation in generateBibFile is not using UTC. I will take a look.

SQR-004 it changed the year but from years ago so why was it wrong yesterday?

womullan:
oh yea pretty crazy - I guess to do with roundtripping in pybtex ...

PR #1123, opened 2024-04-26
 ACM fix the es version

womullan:


etc/glossarydefs_es.csv
```diff
@@ -19,7 +19,7 @@ ACCS,Sistema auxiliar de control de cámaras,
 ACGIH,Conferencia Estadounidense de Higienistas Industriales Gubernamentales,
 ACI,Infraestructura céntrica de aplicaciones,
 ACID,"Atomicidad, Consistencia, Aislamiento y Durabilidad",
-ACM$,Servicio de gestión de fondos de subsidios,
+ACM,Servicio de gestión de fondos de subsidios,
```
plazas, line 22: If we are translating only “ACM” from English, I would think “Gestión de Fondos de Subsidios” o “Gestión de Fondos de Becas” would work.

womullan: Are you inviting any other commenters ?

PR #1171, opened 2024-06-26
 ACP

womullan:


etc/static_entries.bib
```diff
@@ -1956,3 +1956,16 @@ @Misc{Document-25351
 note =          {Vera C. Rubin Observatory Document-25351},
   handle =       {Document-25351},
 }
+
+@Misc{ACP,
```
timj, line 1960: How come this document doesn't get a handle? Shouldn't it at least be a DocuShare document with restricted access? We can't have documents listed that it's impossible for even people on the project to find.

womullan: its a SLAC document who knows ..

timj: For the record, I disagree with this approach.

PR #1210, opened 2024-08-19
convert bibfile generator to algolia 4

womullan:


requirements.txt
```diff
@@ -1,11 +1,12 @@
-algoliasearch<4
 pybtex
 pytest
 pyyaml
 google-api-python-client
 google-auth-httplib2
 google-auth-oauthlib
 oauth2client
+algoliasearch
```
timj, line 8: Can you move this line up to the top so that the diff is less confusing? (and also squash all your commits so you don't have the .bib change in the commit history which you then back out)

PR #1221, opened 2024-08-30
 add ascom author style

womullan:


bin/db2authors.py
```diff
@@ -235,7 +243,10 @@ def get_initials(initials):
     indexOutput.append(rf"%\aindex{{{surname},{justInitials}}}")
 
     if buffer_authors:
-        authOutput.append(author_form.format(initials, surname, affilAuth))
+        if args.mode == "ascom":
+            authOutput.append(author_form.format(affilAuth, initials, surname))
```
timj, line 247: The code could be simplified a lot if we put key names in the format strings and then used kwargs with `format`. eg

```python
>>> f = "affiliation: {affil} author: {author}"
>>> f.format(author="Wil", affil="Rubin")
'affiliation: Rubin author: Wil'
```
Then you wouldn't need to have the if statement here or earlier because the order in the `format` method wouldn't matter anyhow.

timj (approved):
One suggestion for significantly simplifying the logic to remove the if statements that solely change the order of author and affiliation in the fomat method.

PR #1266, opened 2024-10-11
 Paulina Venegas

womullan:


PaulinaLSST (approved):
Hi, thank you for adding me. I will confirm my affiliation with Eric, but in principle is RubinObsC.

PR #1290, opened 2024-11-13
Add BPM.

ktlim:


womullan (changes requested):
can you add spanish also ? (trying to keep it complete)  
Mapa de píxeles defectuoso

ktlim:
@womullan OK, done (and some syntax problems fixed).

etc/glossarydefs_es.csv
```diff
@@ -126,7 +127,7 @@ CA,Cuenta de control (o costo),
 CAA,Comité de Astronomía y Astrofísica (NSF),
 CADC,Centro Canadiense de Datos Astronómicos,
 CALTECH,Instituto Tecnológico de California,
-CAM,CÁMara,DM
```
womullan, line 129: If you remove the tag from here it will break the translation - CAM is overloaded and the TAG breaks the confusion of which translation to use. SO either we put CAM on both or with put DM on both it can not be blank - most docs use DM as default and usually we will want Camera not Control Account Manager ..   I would be ok tagging Camera at "Gen" and Control Account Manager as "PMO" but it must be consistent in both english and spanish files

etc/glossarydefs_es.csv
```diff
@@ -814,7 +815,7 @@ OSHA,Administración de Seguridad y Salud Ocupacional de los EE. UU.,
 OSI,interconexión de sistemas abiertos,
 OSPF,"Open Short Path First [«Abrir el camino más corto primero», protocolo de red]",
 OSPL,OpenSplice DDS - sistema de mensajería subyacente para SAL,
-OSS,Especificaciones del sistema de observatorios; LSE-30,DM
```
womullan, line 817: same issue - tags have to match

womullan (commented):
thanks for catching the comma;s .. i guess i should have a check for that like i do in english

etc/glossarydefs_es.csv
ktlim: Sounds like we need a linter to make sure tags match up and that identical items have different tags...

etc/glossarydefs_es.csv
ktlim: Reverted these.

etc/glossarydefs_es.csv
ktlim: Oh, or did it check and I missed the failure?

womullan (approved):

PR #1293, opened 2024-11-14
 intentional fail on translation tag mismatch

womullan:


bin/generateAcronyms.py
```diff
@@ -798,8 +798,10 @@ def dump_gls(filename, out_file):
                                 trans = transm[tags]
                             else:
                                 print(
-                                    f"Warning: {tags} not in {transm.keys()}  for {acr} - translation missing"
+                                    f"Warning: {tags} not in {transm.keys()}  for"
```
ktlim, line 801: If you're going to exit with an error, this shouldn't be a `Warning`.
```suggestion
                                    f"Error: {tags} not in {transm.keys()}  for"
```

ktlim (approved):

PR #1354, opened 2025-02-11
Fix "affil" keywords for NOIRLab Chile entries

jonathansick:


timj:
At some point I will get around to switching to a pydantic model for validating the entire thing...

PR #1362, opened 2025-02-19
Add orcID and affiliation for Orion Eiger

eigerx:


womullan (approved):
ok - I think no one ever filled in a altaffil ... so lets see if that works

timj:
I think altaffil might only be used in aastex.

PR #1383, opened 2025-03-15
 AAS7 - add email to the output of db2authors

womullan:


timj:
I have to work out a real plan for handling author emails in authordb.yaml since I imagine no-one wants their full plain email in the yaml for easy scraping.

womullan:
i guess we could encrypt and dycrypt with a secret .. will make edit hard ..

timj:
I was going to do something like `tjenness@affil` and then put the `noirlab.edu` or `slac.stanford.edu` or whatever up the top with the affiliations

PR #1385, opened 2025-03-16
Add my email 

leannep:


etc/authordb.yaml
```diff
@@ -1811,6 +1811,7 @@ authors:
     initials: Leanne P.
     name: Guy
     orcid: 0000-0003-0800-8755
+    email: leanne.guy@noirlab.edu
```
timj, line 1814: I will change this when I implement a system that will not make everyone's emails public and scrape-able -- I don't think everyone is going to like their email to be fully formed and public like this.

womullan: mine is in there also .. since long ago.

PR #1397, opened 2025-03-27
DM-49739: Add emails section to authordb

timj:
This contains email domains for affiliations.

Add some email information to author entries.

### DIFF
--- a/bin/db2authors.py
+++ b/bin/db2authors.py
@@ -128,6 +128,9 @@ authorinfo = authordb["authors"]
 affil = authordb["affiliations"]
 affilset = []  # it will be a set but I want index() which is supported in list
 
+# Email domains, keyed on affiliation.
+emails = authordb["emails"]
+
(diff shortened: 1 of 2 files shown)


timj:
@womullan so I have started off with the hack of least resistance. A distinct emails section in authordb and then some code to construct the email from affiliation key.

I added a few emails to the file for testing (from random tech notes I had lying around).

Ideally I would do something like:

```yaml
affiliations:
  RubinObs:
    address: 950 N Cherry Ave
    domain: lsst.org
```
But that's a very big change to the YAML layout and also code changes in db2authors. That's the sort of reorganization I would want to only attempt when I had pydantic models in place. I can save that for the proposed reorganization of the code. This change also won't break the existing code that @jonathansick has for reading this file.

womullan (approved):
looks good - checked out the branch and also ran it on a doc without problem

PR #1415, opened 2025-04-10
Update authordb kelvinls entry to match initials

leeskelvin:


etc/authordb.yaml
```diff
@@ -2304,7 +2304,7 @@ authors:
     initials: Heather
     name: Kelly
     orcid: null
-  kelvinl:
+  kelvinls:
```
timj, line 2307: @leeskelvin  Changing your author ID is dangerous because if you've used it in any other tech note and someone updates to the new author db you will become detached from the history of all your papers.

leeskelvin: Got it, I didn't realize this was a big change. I'll revert the merge.

leeskelvin: Reverted on #1416.

PR #1429, opened 2025-04-23
Fix orcid for caplarn

jonathansick:
I found out the ORCiD for Neven was accidentally copied an pasted from carlinjl. This is the corrected ORCiD.

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -965,7 +965,7 @@ authors:
     altaffil: []
     initials: Neven
     name: Caplar
-    orcid: 0000-0002-3936-9628
+    orcid: 0000-0003-3287-5250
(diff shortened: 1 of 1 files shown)


etc/authordb.yaml
```diff
@@ -965,7 +965,7 @@ authors:
     altaffil: []
     initials: Neven
     name: Caplar
-    orcid: 0000-0002-3936-9628
```
timj, line 968: Turns out that is @jeffcarlin 's ORCID...

timj:
Am I guessing that your pydantic validation spotted the duplicate?

jonathansick:
@timj actually my ingest pipeline with SQLAlchemy ;)

PR #1445, opened 2025-05-07
Update authordb.yaml

Onoddil:
Added SITCOMTN-159 authors and affiliation.


etc/authordb.yaml
```diff
@@ -3031,6 +3032,13 @@ authors:
     initials: Gautham
     name: Narayan
     orcid: null
+  naylor:
```
timj, line 3035: Please use a key of `naylort` to be consistent and `wilsontj` for you below.

I assume Tim does have an ORCID (not important for internal tech notes but will be for other papers)

timj: Looks like it might be 0000-0002-0506-8501

timj: Bonus points if you add Exeter email domain to the email section and then add an `email: timn` to this. (which only matters when AASTeX documents are in play)

PR #1449, opened 2025-05-09
Add Karlo Mrakovcic to authordb

kmrakovcic:
Added Karlo Mrakovcic affiliation

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -329,6 +329,7 @@ emails:
   Stanford: stanford.edu
   Washington-Dirac: uw.edu
   Washington: uw.edu
+  URijeka: uniri.hr
 authors:
   DMPipelines:
     affil:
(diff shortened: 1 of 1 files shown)


etc/authordb.yaml
```diff
@@ -2976,6 +2976,13 @@ authors:
     initials: Christopher B.
     name: Morrison
     orcid: null
+  mrakovcick:
```
timj, line 2979: Can you add the email domain to the emails section up top and then an `email: your_email_user` to this section after orcid? We are trying to get email information into the file (but without putting in explicit emails) since AAS now require it.

kmrakovcic: I think I did what you asked, added URijeka domain into email and then just the part of my email that comes before @ right after the ORCHID info. Hope I understood you correctly.

timj: Perfect. Thanks.

PR #1450, opened 2025-05-09
DM-50816: New implementation of db2authors

timj:
Complete rewrite using per-target classes.

There is still some duplication but at least it's now clearer what is happening and has fixed numerous output bugs.

@jonathansick does not use any of your documenteer classes but hopefully would not be hard to migrate. Still wanting to have minimal dependencies for the moment.

* Fixes output bug in arxiv
* Fixes duplication of affiliation in ascom and spie outputs (and SPIE now uses letters rather than numbers).
* Fixes issue with lsstdoc with "," vs "and".

It's clear that the parsing of affiliation information is not clever enough -- used by ADASS and ASCOM. The best fix would be to change the affiliations in authordb.yaml to explicitly separate the content by institute, city, country, post code etc.

I will rename it to db2authors.py after review (it's easier to see what's there if it's not a diff).

### DIFF
--- a/bin/db2authors.py
+++ b/bin/db2authors.py
@@ -16,335 +16,507 @@ This program requires the "yaml" package to be installed.
 """
 
 import argparse
+import dataclasses
 import os
 import os.path
 import re
+import string
 import sys
+from typing import Any, Self
 
 import yaml
 
-# Set to True to write a comma separated list of authors
-WRITE_CSV = False
 
-# There is a file listing all the authors and a file mapping
-# those authors to full names and affiliations
+def latex2text(latex: str) -> str:
+    """Convert a LaTeX string into a plain text string.
 
-# This is the author list. It's a yaml file with authorID that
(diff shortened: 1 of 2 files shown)


bin/db2authors2.py
```diff
@@ -0,0 +1,496 @@
...
+
+        affiliations = []
+        for affil, number in affil_to_number.items():
+            affiliations.append(rf"\affil[{chars[number - 1]}]{{{affil}}}")
+
+        return self.get_header() + "\n".join(authors + affiliations)
+
+
+class WebOfC(AuthorTextGenerator):
+    """WebOfC generator."""
```
womullan, line 312: not familiar with webofc

womullan (approved):
chmod +x db2authors2.py .. or when you rename it

bin/db2authors2.py
timj: It's the one used by CHEP papers.

PR #1457, opened 2025-05-14
 pkg_resource fix

womullan:


requirements.txt
```diff
@@ -1,3 +1,4 @@
+setuptools
```
timj, line 1: For the other actions I put the setuptools in the action itself because the fact that it needs setuptools is an annoying bug in those packages that are yet to be modernized, but this will work for the moment.

PR #1464, opened 2025-05-20
Add type annotations to python scripts

timj:
This adds type annotations to the python code. The generateAcronyms was a bit tricky to clean up but I think I did it without breaking anything but the code is tricky to follow so you might want a closer look at that. It does seem like things would be a little easier to understand if there was a named tuple for the `tuple[str, str]` part.


womullan (approved):
ok thanks for the typing - runs ok for me - i added PIFF

PR #1477, opened 2025-06-05
Add CSV output to aouthrs and put it in docs

womullan:


bin/db2authors.py
```diff
@@ -132,6 +131,9 @@ def from_authordb(cls, authordb: dict[str, Any]) -> Self:
             authors=authordb["authors"],
         )
 
+    def get_authors(self) -> dict[str, dict]:
```
timj, line 134: Ah. I meant have it return `get_authors(self) -> dict[str, Author]`

This isn't return author anything, it's returning the internal representation but I don't think you are using that.

I think you want

```python
def get_author_ids(self) -> list[str]:
    return list(self._authors)
```
Then that ties in with your usage below.

bin/db2authors.py
```diff
@@ -477,6 +479,21 @@ def generate(self) -> str:
         )
 
 
+def dump_csv(factory: AuthorFactory) -> None:
+    """Generate CSV of ALL authors for easier lookup of ID .
+    Authorid, Name, Institution id
+    put this in authors.csv
+    """
+    authors: dict[str, dict] = factory.get_authors()
```
timj, line 487:
```suggestion
    author_ids = factory.get_author_ids()
```

timj (approved):
Thanks. Looks okay. Some minor comments and a misunderstanding cleared up.

bin/db2authors.py
```diff
@@ -477,6 +479,21 @@ def generate(self) -> str:
...
+    with open("authors.csv", "w") as outf:
+        print("Rubin ID, Name, Affiliation(s)", file=outf)
+        for authorid in authors.keys():
+            author = factory.get_author(authorid)
+            line = f'{authorid},{author.full_name},"{author.affiliations}"'
+            print(line, file=outf)
+        outf.close()
```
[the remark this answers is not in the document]
timj: 😄 the entire point of the context manager is to auto close, even if there is an error.

PR #1491, opened 2025-06-19
 form processor - not sure i like the new author output

womullan:


bin/makeAuthorListsFromGoogle.py
```diff
@@ -0,0 +1,250 @@
...
+
+# If modifying these scopes, delete your previously saved credentials
+# at ~/.credentials/sheets.googleapis.com-python-quickstart.json
+SCOPES = ["https://www.googleapis.com/auth/spreadsheets.readonly"]
+CLIENT_SECRET_FILE = "client_secret.json"
+APPLICATION_NAME = "Process Authors Google Sheet"
+
+
+@dataclasses.dataclass(frozen=True)
+class AuthorY:
```
timj, line 48: Isn't this dataclass now identical to Author though?

We could easily make altaffil have a defaulted factory for an empty list. You likely aren't asking for altaffil in your form..

bin/makeAuthorListsFromGoogle.py
```diff
@@ -0,0 +1,250 @@
...
+            if len(id) == 0:  # may be an update
+                id = row[6]
+            if len(id) < 3:
+                initials = get_initials(row[8])
+                id = f"{row[7]}{initials}"  # last name initial
+                id = id.lower()
+                # loaded the authorids from suthordb and check ..
+                if id in authors:
+                    print("Perhaps check  clash - author {id} - {row[5]}, {row[6]} ")
+                    clash.append(id)
```
timj, line 155: The code continues even if it thinks there is a clash?

bin/makeAuthorListsFromGoogle.py
```diff
@@ -0,0 +1,250 @@
...
+        factory = AuthorFactory.from_authordb(authordb)
+        authors = factory.get_author_ids()
+        affils = factory.get_affiliation_ids()
+
+        for row in values:
+            id = row[5]
+            if len(id) == 0:  # may be an update
+                id = row[6]
+            if len(id) < 3:
+                initials = get_initials(row[8])
```
timj, line 149: Main problem here is that (1) people don't always put initials in and (2) some people have made authorIDs that didn't quite follow the rules and so don't have the middle initial in (and it's difficult to change now).

bin/makeAuthorListsFromGoogle.py
```diff
@@ -0,0 +1,250 @@
...
+        dbfile = os.path.normpath(os.path.join(exedir, os.path.pardir, "etc", "authordb.yaml"))
+        with open(dbfile) as fh:
+            authordb = yaml.safe_load(fh)
+
+        factory = AuthorFactory.from_authordb(authordb)
+        authors = factory.get_author_ids()
+        affils = factory.get_affiliation_ids()
+
+        for row in values:
+            id = row[5]
```
timj, line 145: I think the code would be much easier to understand if you made a NamedTuple from the row.

```python
>>> class Row(NamedTuple):
...     id: str
...     name: str
...     affil: str
... 
>>> Row("id", "name", "affil")
Row(id='id', name='name', affil='affil')
>>> row = Row("id", "name", "affil")
>>> row.name
'name'
```

Then you don't have to remember the indexes for the columns.

bin/makeAuthorListsFromGoogle.py
```diff
@@ -0,0 +1,250 @@
...
+                        affil = affilid
+                        affilid = get_initials(affil)
+                        newaffils[affilid] = affil
+
+                # we have a name so we need to gather the rest.
+                orcid = row[10]
+                if len(orcid) == 0:
+                    orcid = None
+                email: str = row[1]
+                if "@" in email:
```
timj, line 178: This assumes that the affiliation default email is correct but without validating it (it won't work for people putting NOIRLab affiliation but lsst.org email). Hope it's okay...

bin/makeAuthorListsFromGoogle.py
```diff
@@ -0,0 +1,250 @@
...
+    count = 0
+    for n in all:
+        count = count + 1
+        initials = f"{initials}{n[0]}"
+        if count > 5:
+            break
+    return initials
+
+
+def wrtie_yaml(name: str, values: Any) -> None:
```
timj, line 105: "wrtie" typo but it is consistent in the file.

timj (approved):
It's fine.

I think Jonathan has a class that represents the native form of the authordb.yaml itself and can probably regenerate it (he does a two stage approach of one model representing the file layout and another model representing the expanded form of it). I didn't use that code in db2authors immediately because I wasn't sure how to handle the pydantic dependency.

It sounds like we will need something like that though.

I also intend to update the affiliation section to at least separate out department, organization, address, and ROR ID (the latter is needed for DOIs) but I need to coordinate that change with Jonathan.

bin/makeAuthorListsFromGoogle.py
womullan: yes the id is included in the authorlist - perhaps auhtor did not know they existed .. we cna not fix the input i guess so we will have to take care of this after before we merge to authodb,,

bin/makeAuthorListsFromGoogle.py
womullan: here i just want to skim whatever they put to estimate an id ..  they have to put at least a name on the form

bin/makeAuthorListsFromGoogle.py
womullan: yes corner cutting -- how important is that email in the authordb ..  i guess i could validate - how are you handling the non standard in the end ?

bin/makeAuthorListsFromGoogle.py
womullan: i guess i dont understand how that works/helps - the form alas has id in two locations depending -- and the column names are long strings - I only use the index once to get the value ..  but i will have a look

bin/makeAuthorListsFromGoogle.py
womullan: ok - now added get domains to authorfactory - check the domain is listed replace the domain with  domainID, if not listed create a new domain - those go in a file to also be merged to authordb

PR #1495, opened 2025-06-23
Remove duplicate EA entry in glossarydefs.csv

jonathansick:
This got added from some CSV churn on the weekend.


womullan (approved):
thanks

PR #1498, opened 2025-06-27
Remove duplicate that is also in ivoa.bib file

leannep:


womullan (approved):
ok - i dont remember adding that  ivoa.bib .!!

PR #1508, opened 2025-06-30
update to authordb w/ Sierra

villarrealas:
Just a quick add of Sierra Villarreal to the author list.

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -4495,6 +4495,7 @@ authors:
     - SLAC
     initials: Antonia S.
     name: Villarreal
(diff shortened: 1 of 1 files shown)


etc/authordb.yaml
```diff
@@ -4490,6 +4490,12 @@ authors:
     initials: Ashley
     name: Villar
     orcid: 0000-0002-5814-4061
+  villarreala:
```
timj, line 4493: CI is failing because you need the altaffil field (it doesn't default)

timj (approved):

PR #1509, opened 2025-06-30
SP-2384: Updates for DP1 

leannep:


etc/glossarydefs.csv
```diff
@@ -691,7 +691,7 @@ LSP,LSST Science Platform (now Rubin Science Platform),LSST DM,,,A
 LSR,LSST System Requirements; LSE-29,LSST DM,,,A
 LSS,Large Scale Structure,Sci,,,A
 LSST,Legacy Survey of Space and Time (formerly Large Synoptic Survey Telescope),Gen,,,A
-LSSTC,LSST Corporation,Adm,,,A
```
womullan, line 694: we should probably keep that .. it was a thing and i is in many docs good to add the new one and change DSFP to use it

womullan (approved):
IN some other case i have put a definitiion like  
"LSST Corporation - obsolete use LSSTDA"
rather than remove

etc/glossarydefs.csv
timj: I agree it should stay since there are old docs around, but possibly as something like "LSST Discovery Alliance (formerly LSST Corporation)".

etc/glossarydefs.csv
```diff
@@ -1527,7 +1527,7 @@ Level 3 Data Product,Deprecated term; see User Generated Data Product,DM,LPM-231
...
-LSST Corporation,"An Arizona 501(c)3 not-for-profit corporation formed in 2003 for the purpose of designing, constructing, and operating the LSST System. During design and development, the Corporation stewarded private funding used for such essential contributions as early site preparation, mirror construction, and early data management system development. During construction, LSSTC will secure private operations funding from international affiliates and play a key role in preparing the scientific community to use the LSST dataset",Adm,,,G
+LSST Corporation,"An Arizona 501(c)3 not-for-profit corporation formed in 2003 for the purpose of designing, constructing, and operating the LSST System. During design and development, the Corporation stewarded private funding used for such essential contributions as early site preparation, mirror construction, and early data management system development. During construction, LSSTDA will secure private operations funding from international affiliates and play a key role in preparing the scientific community to use the LSST dataset",Adm,,,G
```
timj, line 1530: Isn't the last sentence here completely incorrect now? That never happened.

timj:
Guthub has got confused in this PR now and thinks there are loads of commits that are already on main

etc/glossarydefs.csv
womullan: historically correct - that was what it was intended for

PR #1514, opened 2025-07-02
DM-49840: Add DP1 dataset DOIs and LSSTComCam DOI to static values

timj:


timj:
This is failing because we have more than 26 bib entries with the same author-year (the DP1 DOIs) and when the bst file decides to go one up from `z` it ends up with `{` and of course everything blows up. Not entirely sure what to do about it other than tweaking a couple of the authors to be something different.

PR #1516, opened 2025-07-03
add kherner

kherner:


etc/authordb.yaml
```diff
@@ -2396,6 +2396,14 @@ authors:
     initials: Richard
     name: Kessler
     orcid: 0000-0003-3221-0419
+  kherner:
```
timj, line 2399: hernerk -- alphabetical

timj (changes requested):

etc/authordb.yaml
kherner: Fixed

PR #1519, opened 2025-07-04
DM-51720: New version of authordb with rich affiliations

timj:
db2authors modified to use it.

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -1,357 +1,2747 @@
 affiliations:
-  AAS: American Astronomical Society
-  Adler: The Adler Planetarium, 1300 South Lakeshore Ave., Chicago, IL 60605, USA
-  ALMA: Atacama Large Millimeter/submillimeter Array, San Pedro de Atacama, Antofagasta, Chile
-  AMNH: Department of Astrophysics, American Museum of Natural History, New York,
-    NY 10028, USA
-  ASTRON: ASTRON, Oude Hoogeveensedijk 4, 7991\,PD, Dwingeloo, The Netherlands
-  AWSS: Amazon Web Services, Seattle, WA 98121, USA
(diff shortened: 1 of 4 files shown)


timj:
I have also done the given_name / family_name migration to be consistent with DataCite.

This does mean we now have model code in authordb.py and db2authors.py and duplication of email domain forming as well, but at least the YAML is easier to understand.

I haven't done the switch to Unicode because that would add a new dependency for db2authors and if we are doing that I might as well use pydantic and move all the model code to authordb.py. I don't understand how we can inject pylatexenc and pydantic into all the github actions for tech notes without having to go in and patch each one because the actions aren't using a reusable component anywhere.

etc/authordb.yaml
```diff
@@ -1,1624 +1,4247 @@
...
+      street: ''
+    department: Department of Astrophysics,
+    email: null
+    institute: American Museum of Natural History
+    ror_id: null
+  ARCSoM:
+    address:
+      city: Sheffield
+      country_code: UK
+      example_expanded: Astrophysics Research Cluster, School of Mathematical and
```
timj, line 44: There is a burden to *requiring* we fill in this section since:

* It can get out of sync with the rest of the address contents.
* It is duplicating content
* If we are asking for this string from people we definitely want to be trying to parse it ourselves because that is not going to work and if we could do that we wouldn't have changed anything --- we want the submitted information to not include this at all but include the individual sections
* Including the university in the address class is inconsistent with the model layout.

I am inclined to delete it completely before merging because of this. @jonathansick  you indicated that you liked it being there.

womullan (approved):
nice

timj:
Merging this without answering all the questions because it's causing confusion and is impossible for me to rebase if someone else edits authordb.yaml.

PR #1523, opened 2025-07-06
add Tianqing Zhang to db

ztq1996:
Add Tianqing Zhang to authorDB.yaml

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -7823,6 +7823,14 @@ authors:
     given_name: Ann
     family_name: Zabludoff
     orcid: 0000-0001-6047-8469
+  zhangt:
+    affil:
+    - Pitts
+    altaffil: []
+    email: tq.zhang@pitt.edu
+    given_name: Tianqing
+    family_name: Zhang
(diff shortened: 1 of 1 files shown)


etc/authordb.yaml
```diff
@@ -7823,6 +7823,14 @@ authors:
     given_name: Ann
     family_name: Zabludoff
     orcid: 0000-0001-6047-8469
+  zhangt:
+    affil:
+    - Pitts
+    altaffil: []
+    email: tq.zhang@pitt.edu
```
timj, line 7830: Rather than making your email public to everyone, what you can do is put  `email: tq.zhang` here and because there is an email defined in the `Pitts` definition above it will correctly work out your real email without it being visible to search engines here.

etc/authordb.yaml
```diff
@@ -7823,6 +7823,14 @@ authors:
...
+    given_name: Tianqing
+    family_name: Zhang
+    orcid: 0000-0002-5596-198X
```
timj, line 7833: Trailing spaces need to be removed.

ztq1996:
Thank you, Tim. Done!

PR #1524, opened 2025-07-07
 a couple of missing email and 2 new affil

womullan:


etc/authordb.yaml
```diff
@@ -11,6 +11,19 @@ affiliations:
...
     ror_id: null
+  AEL:
+    address:
+      city: Kiruna
+      country_code: Sweden
+      example_expanded: Asteroid Engineering Laboratory, Lule\xE5 University of
+         Technology, Box 848, SE-981 28 Kiruna, Sweden
+      postcode: 'SE-981 28'
+      state: null
+      street: Lule\o{a} University of Technology, Box 848
```
timj, line 22: This is the university which should be in the institute field below and the institute is the department...

PR #1532, opened 2025-07-10
updated affiliation

psferguson:


psferguson:
> Fly-by review: You might want to include UW Madison to `altaffil`.

based on this slack message: https://rubin-obs.slack.com/archives/C08FVN6U4V7/p1751907744602969?thread_ts=1751903233.909999&cid=C08FVN6U4V7

"Altaffil is not their second affiliation, it’s for things like “NASA Fellow” or “Deceased”"

And I think I am no longer affiliated with UW Madison

timj:
Bonus points if you can fill in the email field. (you just need the part before the `@uw.edu` because it knows the email domain from your affiliation. AAS require emails now for all authors.

(and yes, multiple affiliations go in the affiliations field and altaffil is as you say).

arunkannawadi:
Right, forgot that `affil` takes a list. Since a good chunk of the work for DP1 happened while you were employed at Madison, I think it'd be fair to list both. But you know your work allocation better.

PR #1541, opened 2025-07-11
 Brown - and more authors

womullan:


etc/authordb.yaml
```diff
@@ -2570,6 +2578,20 @@ affiliations:
     email: null
     institute: UC Santa Cruz
     ror_id: null
+  UCS:
```
timj, line 2581: Haven't you put in twice UCS and USC as the same place?

etc/authordb.yaml
```diff
@@ -5122,6 +5175,14 @@ authors:
     family_name: Jennings
     given_name: Toby C.
     orcid: 0009-0000-7029-5690
+  jeremiea:
+    affil:
+    - IN2P3-LAPP
+    altaffil: []
+    email: andrea
+    family_name: JEREMIE
```
timj, line 5183: Upper cased.

etc/authordb.yaml
timj: Removed this entry.

PR #1548, opened 2025-07-12
 more DP1 authors, affiliations

womullan:


etc/authordb.yaml
```diff
@@ -234,6 +234,19 @@ affiliations:
     email: null
     institute: University of California, Berkeley
     ror_id: 01an7q238
+  Berkley-Physics:
```
timj, line 237: Spelling error

etc/authordb.yaml
```diff
@@ -5258,13 +5271,12 @@ authors:
     orcid: null
   kahnsm:
     affil:
-    - RubinObs
-    - SLAC-Kavli
+    - Berkley-Physics
     altaffil: []
-    email: null
+    email: stevkahn
```
timj, line 5276: Odd that they dropped the e

womullan: apparenlty correct htough

PR #1554, opened 2025-07-14
Remove incorrect orcid for sahaa

jonathansick:
Based on https://orcid.org/0000-0002-5091-0470 this orcid should belong
to schmidts

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -7124,7 +7124,7 @@ authors:
     email: null
     family_name: Saha
     given_name: Abhijit
-    orcid: null
+    orcid: 0000-0002-6839-4881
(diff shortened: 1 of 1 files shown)


etc/authordb.yaml
```diff
@@ -7124,7 +7124,7 @@ authors:
     email: null
     family_name: Saha
     given_name: Abhijit
-    orcid: 0000-0002-5091-0470
+    orcid: null
```
timj, line 7127: I have no idea how this happened since it's been in there since 2019 (from me).

timj: ADS tells me his ORCID should be 0000-0002-6839-4881

PR #1558, opened 2025-07-14
 2 more authors

womullan:


etc/authordb.yaml
```diff
@@ -3951,11 +3959,11 @@ authors:
     orcid: 0000-0001-5576-8189
   constanzocordovaje:
     affil:
-    - RubinObsC
+    - RubinObs
```
timj, line 3962: Julio isn't in Chile?

womullan: yea i explained - he got it wrong

PR #1562, opened 2025-07-14
Update Colin Orion Chandler authordb

wilsonbb:
Update Colin Orion Chandler's information in the author db

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -1624,14 +1624,14 @@ affiliations:
     address:
       city: Tucson
       country_code: USA
-      example_expanded: LSST Interdisciplinary Network for Collaboration and 
+      example_expanded: LSST Interdisciplinary Network for Collaboration and
         Computing, Tucson, USA
       postcode: '85721'
       state: AZ
       street: 933 North Cherry Avenue
     department: null
     email: null
-    institute: LINCC
+    institute: LSST Interdisciplinary Network for Collaboration and Computing
     ror_id: null
   Longhorn:
     address:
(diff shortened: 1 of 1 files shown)


etc/authordb.yaml
```diff
@@ -1620,6 +1620,19 @@ affiliations:
...
+      city: Tucson
+      country_code: USA
+      example_expanded: LSST Interdisciplinary Network for Collaboration and 
+        Computing, Tucson, USA
+      postcode: '85721'
+      state: AZ
+      street: 933 North Cherry Avenue
+    department: null
+    email: null
+    institute: LINCC
```
timj, line 1634:
```suggestion
    institute: LSST Interdisciplinary Network for Collaboration and  Computing
```
I think this should be the same string you use in the example (although there is no problem adding "(LINCC)" at the end).

etc/authordb.yaml
```diff
@@ -3774,9 +3787,9 @@ authors:
     orcid: null
   chandlerc:
     affil:
-    - NAU
-    altaffil: []
-    email: null
+    - LINCC
+    altaffil: [Washington, NAU]
+    email: coc123@uw.edu
```
timj, line 3792:
```suggestion
    email: coc123@Washington
```
(so you don't have your email fully obvious).

etc/authordb.yaml
```diff
@@ -3774,9 +3787,9 @@ authors:
     orcid: null
   chandlerc:
     affil:
-    - NAU
-    altaffil: []
-    email: null
+    - LINCC
+    altaffil: [Washington, NAU]
```
timj, line 3791: Altaffil is the `\altaffiliation` field in AASTeX and is for things like "Author is deceased". In this case you want to put these two in the `affil` section like:
```yaml
affil:
- LINCC
- Washington
- NAU
```

etc/authordb.yaml
wilsonbb: Done.

etc/authordb.yaml
wilsonbb: Done.

etc/authordb.yaml
wilsonbb: Done.

PR #1570, opened 2025-07-16
Switch leget from Princeton to Princeton-Astro

PFLeget:


etc/authordb.yaml
```diff
@@ -5802,7 +5802,7 @@ authors:
     orcid: null
   legetp:
     affil:
-    - Princeton
+    - Princeton-Astro
     altaffil: []
     email: leget@Princeton-Astro
```
timj, line 5807: You can delete the @Princeton-Astro text as well then.

PR #1598, opened 2025-07-22
 Add single author tex output and put it in each line of CSV

womullan:


timj:
Did you mean to commit that glossary tex file?

I was thinking that it might be better for things like ADASS if instead of adding a new method for single author (which wouldn't pick up the ADASS extras) we instead modified the generate() method to allow a single author override that would be passed in as an optional parameter and be used instead of `self.authors`?

timj:
Ah yes. A header flag seems okay to me.

womullan:
ok with header boolean and all generators modified to use it

bin/db2authors.py
```diff
@@ -345,11 +354,11 @@ def _generate_paren_text(self, author: Author) -> str:
         """
         return f"[{author.orcid}]" if author.orcid else ""
 
-    def generate(self) -> str:
+    def generate(self, header: bool = True) -> str:
         """Generate AASTeX format."""
         lines = []
+        lines.append("")
```
timj, line 360: I wanted a blank line between each author entry though...

bin/db2authors.py
```diff
@@ -624,11 +633,17 @@ def dump_csvall(factory: AuthorFactory) -> None:
...
-            affils = " | ".join(a.get_full_address_with_institute() for a in author.affiliations)
-            line = f'{authorid},{latex2text(author.full_name)},"{latex2text(affils)}"'
+            aas7_generator = AASTeX7([author])
+            # people seeing the full affiliation copied it so putting IDs
+            affils = " / ".join(factory.get_affiliation_id(a) for a in author.affiliations)
+            # Generate aas7 string using the AASTeX7 generator
+            aas7_text = aas7_generator.generate(header=False)
+            # the email should not be plain ..
+            aas7_text = aas7_text.replace("@", " AT ")
+            line = f'{authorid},{latex2text(author.full_name)},{affils},"{aas7_text}"'
```
timj, line 646: Does this work with the `, Jr` name?  Consider using the standard `csv` package for writing the CSV so you don't have to worry about quoting.

womullan: confirmed - even in the current live version this is an issue

PR #1605, opened 2025-07-24
Update authordb

combet:
Add missing authors for SITCOMTN-161


timj:
PS The error is a trailing space in the yaml file.

PR #1622, opened 2025-07-28
 some more DP1 updates, emails  vauvher emerson updates

womullan:


etc/authordb.yaml
```diff
@@ -3931,9 +3931,9 @@ authors:
     orcid: 0000-0002-3925-7802
   charlesemersong:
     affil:
-    - RubinObs
+    - AURA-T
```
timj, line 3934: I'm pretty sure we want Glenaver to be Rubin don't we?

womullan: she choose this but yea i can flip it back

PR #1657, opened 2025-08-09
Normalize unicode author names before fixing multi-surname names

timj:
We were running into trouble with `\'e` vs `\'{e}` and it's easier to use the unicode version for replacement and then convert back to standard latex form.


womullan (approved):
Good to make them all the same.

PR #1710, opened 2025-08-26
final homogenization of French institutions to make Johan happy

bregeon:
adjustments to authors.db


timj (approved):
Thanks for sorting these out. It's incredible work 😄

PR #1742, opened 2025-08-29
 Use orclink for ASCOM since [orcid] does not work

womullan:
will have to update the template to include \usepackage{orcidlink}  which is in standard texlive.  We could  use this in our docs as well if we wanted.


womullan:
no they do not say at all how to include orcid - but this seems to work .

PR #1745, opened 2025-08-30
 Do not include in etc and local

womullan:


bin/db2authors.py
```diff
@@ -751,6 +752,39 @@ def dump_csvall(factory: AuthorFactory) -> None:
...
+        Same format as authors.yaml
+
+    Returns
+    -------
+    list[str]
+        A list of author IDs (strings)
+    """
+    authorids: list[str] = []
+    with open(donotinclude, encoding="utf-8") as f:
+        ids = yaml.safe_load(f)
```
timj, line 773: Maybe do:
```python
authorids = yaml.safe_load(f)
```
? Do we need `ids` variable?

bin/db2authors.py
```diff
@@ -751,6 +752,39 @@ def dump_csvall(factory: AuthorFactory) -> None:
...
 
 
+def load_dni(donotinclude: str) -> list[str]:
+    """
+    Load  the standard 'do not include' file and return a list of author IDs.
+    Look for a local one also and add that.
+
+    Parameters
+    ----------
+    donotinclude : str
```
timj, line 762:
```suggestion
    donotinclude : `str`
```

timj (approved):
Looks good although I think a set() would simplify things.

bin/db2authors.py
```diff
@@ -751,6 +752,39 @@ def dump_csvall(factory: AuthorFactory) -> None:
...
+    -------
+    list[str]
+        A list of author IDs (strings)
+    """
+    authorids: list[str] = []
+    with open(donotinclude, encoding="utf-8") as f:
+        ids = yaml.safe_load(f)
+    if ids:
+        authorids = ids
+    if Path("dni.yaml").exists():  # local per doc
```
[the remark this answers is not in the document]
womullan: it seems simlar to me but i used the path.open

bin/db2authors.py
```diff
@@ -751,6 +752,38 @@ def dump_csvall(factory: AuthorFactory) -> None:
...
+        A list of author IDs (strings)
+    """
+    with open(donotinclude, encoding="utf-8") as f:
+        authorids = yaml.safe_load(f)
+    dnip = Path("dni.yaml")
+    if dnip.exists():  # local per doc
+        with dnip.open(encoding="utf-8") as f:
+            dni_local: set[str] = yaml.safe_load(f)
+            if authorids:
+                authorids.update(dni_local)
```
timj, line 778: Does this work? authorids is a `list` still (from line 772)  and declaring dni_local to be a set doesn't make it a set when `safe_load` is returning a list.

bin/db2authors.py
```diff
@@ -751,6 +752,38 @@ def dump_csvall(factory: AuthorFactory) -> None:
...
+            else:
```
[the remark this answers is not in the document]
timj: (but a set would be better because you use `in` below I think).

PR #1764, opened 2025-09-05
DM-52431: Add support for AAS author submission CSV output

timj:


timj:
The first column is for the corresponding author, which isn't encoded in the authors.yaml. What I could do is support an annotation such as:
```yaml
- jennesst*
- omullanew
```
and detect the `*` to mean corresponding author. Would need to discuss with Jonathan (since he's considering switching the lander author detection to use authors.yaml.

womullan:
yea i was wondering about that corresponding author - i guess in the signup form i would also have to add a question to catch that

timj:
We can't be asking people on the google form whether they think they are the corresponding author though. That is going to lead to lots of confusion given that we know that the person leading the paper already knows the answer.

womullan:
yea so perhaps another file ? corresponding-author.yaml  can there be more than one ?

PR #1782, opened 2025-09-14
Automatic authordb  update

womullan:


timj (changes requested):
We still do not need this entry because the author already exists in the line below

etc/authordb.yaml
```diff
@@ -4389,6 +4389,14 @@ authors:
     family_name: Cockrum
     given_name: Joseph J.
     orcid: null
+  cohentanugi,johanncj:
```
timj, line 4392: Is the system continuing to try the same entry?

PR #1784, opened 2025-09-15
Automatic authordb  update

womullan:


womullan:
no - if the skip is updated it will not try that row again.Its part of the PR
It is in my OTHER PR

PR #1975, opened 2026-04-01
Automatic authordb  update

womullan:


etc/authordb.yaml
```diff
@@ -3088,6 +3088,19 @@ affiliations:
     email: null
     institute: Universidad de los Andes
     ror_id: 02mhbdp94
+  UoWDoA:
```
timj, line 3091: We already have a UW affiliation.

PR #1985, opened 2026-04-13
DM-54622: Reorganize Rubin affiliations for construction

timj:
This pull request:

* Makes RubinObs and RubinObsC the primary Rubin AURA affiliation.
* Converts NOIRLab Rubin staff to use noirlab.edu email
* People who are not on staff Slack and not in the NOIRLab contacts database, explicitly mark them as construction affiliations (people who leave project but who are builders can retain a construction affiliation if they are no longer working in astronomy)
* Added a RubinObsSLAC affiliation for SLAC staff matching the RubinObs NOIRLab approach.
* Added RubinFundedSLAC and RubinFundedAURA affiliations for staff at other locations who are funded by the observatory  which can be used as a second affiliation.

cc/ @drphilmarshall for comment on the SLAC entries specifically (since you requested these).

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -1012,7 +1012,7 @@ affiliations:
       street: 670 N.\ A’ohoku Place
     department: null
     email: null
-    institute: International Gemini Observatory/NSF NOIRLab
+    institute: International Gemini Observatory / NSF NOIRLab
     ror_id: 0141j1w94
   GeminiS:
     address:
@@ -1025,7 +1025,7 @@ affiliations:
       street: Casilla 603
     department: null
     email: null
-    institute: International Gemini Observatory/NSF NOIRLab
+    institute: International Gemini Observatory / NSF NOIRLab
     ror_id: 02wtj7z45
   GeorgiaState:
     address:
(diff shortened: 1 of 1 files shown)


etc/authordb.yaml
```diff
@@ -2336,58 +2336,109 @@ affiliations:
...
+      street: 2575 Sand Hill Rd.
+    department: NSF-DOE Vera C.\ Rubin Observatory
+    email: null
+    institute: SLAC/DOE
+    ror_id: 05gzmn429
+  RubinObs:
+    address:
+      city: Tucson
+      country_code: USA
+      example_expanded: NSF-DOE Vera C.\ Rubin Observatory / NSF NOIRLab, 950 N.\
```
timj, line 2368: @rblum5 also, wouldn't a comma make more sense than a '/' to separate things?

womullan (approved):
major updates ! Thanks
Do we have to do through another day and put in the second affiliation for people ? Plus that is, at leas tin theory,  optional for them ..

timj:
I was going to do the addition of the Rubin funded affiliations in a second pass.

I don't think they are really optional for any of the papers where people are using the authordb...

timj:
Manuel Gomez has requested:

* Change family name to "Gomez-Jimenez"
* Add his ORCID  0009-0006-5048-0972
* I think we also need to make sure he is RubinObsC affiliation.

PR #1994, opened 2026-05-04
Update authordb.yaml

jrob93:
Adding my author information for use in http://rtn-121.lsst.io/

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -8159,6 +8159,14 @@ authors:
     family_name: Robertson
     given_name: Brant E.
     orcid: 0000-0002-4271-0364
+  robinsonje:
+    affil:
+    - EdinburghIA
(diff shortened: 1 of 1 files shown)


etc/authordb.yaml
```diff
@@ -8138,6 +8138,14 @@ authors:
     family_name: Robertson
     given_name: Brant E.
     orcid: 0000-0002-4271-0364
+  robinsonje:
+    affil:
+    - EdinburghIA
+    altaffil: []
+    email: james.robinson@ed.ac.uk
```
timj: Don't use a full email address. If the email domain for EingburghIA is correct then you don't need anything other than `james.robinson`.

jrob93: Apologies, thanks for catching that. Should be resolved now

PR #1996, opened 2026-05-05
U/garrethmartin/update authordb

garrethmartin:
Updated the authordb.yaml with my affiliations

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -1987,6 +1987,19 @@ affiliations:
     email: noirlab.edu
     institute: NSF NOIRLab
     ror_id: 03zmsge54
+  Nottingham:
+    address:
+      city: Nottingham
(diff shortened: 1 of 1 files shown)


etc/authordb.yaml
```diff
@@ -1987,6 +1987,19 @@ affiliations:
...
+    address:
+      city: Nottingham
+      country_code: United Kingdom
+      example_expanded: School of Physics \& Astronomy, University of Nottingham, University Park,
+        Nottingham NG7 2RD, UK
+      postcode: NG7 2RD
+      state: ''
+      street: University Park
+    department: School of Physics \& Astronomy
+    email: null
```
timj: Please add the email domain here. AAS now insist on emails.

timj (approved):
Looks okay to me if the email is fixed.

PR #1997, opened 2026-05-06
Automatic authordb  update

womullan:


etc/authordb.yaml
```diff
@@ -2505,6 +2505,18 @@ affiliations:
...
     institute: Space Telescope Science Institute
     ror_id: 036f5mx38
+  SUoTMV:
+    address:
+      city: ''
+      country_code: Australia
+      example_expanded: Swinburne University of Technology, Melbourne, VIC, Australia
+      postcode: ''
+      state: null
+      street: Melbourne, VIC
```
timj: This seems like it's not a street.

etc/authordb.yaml
```diff
@@ -3274,6 +3286,12 @@ affiliations:
     email: null
     institute: soZen Inc.
     ror_id: null
+  swin:
```
timj: This should not exist. I don't think it's needed.

etc/authordb.yaml
```diff
@@ -2505,6 +2505,18 @@ affiliations:
     email: stsci.edu
     institute: Space Telescope Science Institute
     ror_id: 036f5mx38
+  SUoTMV:
+    address:
+      city: ''
+      country_code: Australia
+      example_expanded: Swinburne University of Technology, Melbourne, VIC, Australia
+      postcode: ''
+      state: null
```
timj: Presumably Victoria goes here.

etc/authordb.yaml
```diff
@@ -2505,6 +2505,18 @@ affiliations:
...
+      street: Melbourne, VIC
+    department: null
+    email: swin.edu.au
+    institute: Swinburne University of Technology
+    ror_id: null
```
timj: Needs a ROR

timj (changes requested):
There are two distinct changes here. One is the sorting and the other is the new affiliation.

PR #2000, opened 2026-05-08
Add 2 new references for DP1 paper

leannep:
- Add Merritt SORCHA reference
- Update DES Y6 paper to ADS citation

### DIFF
--- a/texmf/bibtex/bib/refs.bib
+++ b/texmf/bibtex/bib/refs.bib
@@ -1,4 +1,4 @@
-i@article{Merritt_2025,
+@article{Merritt_2025,
 doi = {10.3847/1538-3881/add3ec},
 url = {https://doi.org/10.3847/1538-3881/add3ec},
 year = {2025},
--- a/texmf/bibtex/bib/refs.bib
+++ b/texmf/bibtex/bib/refs.bib
@@ -1,18 +1,3 @@
-@article{Merritt_2025,
-doi = {10.3847/1538-3881/add3ec},
-url = {https://doi.org/10.3847/1538-3881/add3ec},
-year = {2025},
-month = {jul},
-publisher = {The American Astronomical Society},
-volume = {170},
-number = {2},
-pages = {100},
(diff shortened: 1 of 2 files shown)


texmf/bibtex/bib/refs.bib
```diff
@@ -1,4 +1,17 @@
-
+i@article{Merritt_2025,
```
timj:
```suggestion
@article{Merritt_2025,
```

Please use the ADS form from https://ui.adsabs.harvard.edu/abs/2025AJ....170..100M/exportcitation and put in refs_ads.bib.

texmf/bibtex/bib/refs_ads.bib
```diff
@@ -6603,21 +6603,23 @@ @ARTICLE{2023PASP..135k5003E
...
-    month = "January",
-    eid = "arXiv:2501.05739",
-    pages = "arXiv:2501.05739",
-    doi = "10.48550/arXiv.2501.05739",
-    archivePrefix = "arXiv",
-    eprint = "2501.05739",
-    primaryClass = "astro-ph.CO",
-    adsurl = "https://ui.adsabs.harvard.edu/abs/2025arXiv250105739B",
-    adsnote = "Provided by the SAO/NASA Astrophysics Data System"
+@ARTICLE{2026ApJS..282...62B,
```
timj, line 6624: This will have to move to its sorted location. We also have to be careful *deleting* entries because if other papers get rebuilt that referred to the arXiv one they break. In the past I've duplicated the entries so included the new text with the old bib key and then added the new bibkey in its right place.

timj (changes requested):

texmf/bibtex/bib/refs.bib
leannep: removed from refs.bib and added ADS form to refs_ads.bib

texmf/bibtex/bib/refs_ads.bib
leannep: ok

timj (approved):
I still worry that we need to keep the old

```
@ARTICLE{2025arXiv250105739B,
```
entry around but with the new published content for it (unfortunately bibtex does not support bib key aliasing).

Remember not to sync the dp1 paper with this new lsst-texmf though since it will break all the author affiliations.

PR #2032, opened 2026-06-09
Add affiliation for Christine Corry

christinecorry:


etc/authordb.yaml
```diff
@@ -4601,6 +4601,14 @@ authors:
     family_name: Corral
     given_name: Luis
     orcid: null
+  corryc:
+    affil:
+    - Berkeley-Astro
+    altaffil: []
+    email: c.corry@berkeley
```
timj: You can just say "c.corry" here and it will work out your default email domain from the "Berkeley-Astro" affiliation.

etc/authordb.yaml
```diff
@@ -4601,6 +4601,14 @@ authors:
     family_name: Corral
     given_name: Luis
     orcid: null
+  corryc:
+    affil:
+    - Berkeley-Astro
```
timj, line 4606: Turns out that:

> Author corryc has email defined but associated affiliation Berkeley-Astro has no email defined

so can you please add the email up to the top of this file at the Berkeley-Astro definition please.

timj:
Thanks for sending your information.

PR #2043, opened 2026-06-22
VUCA and several refs (non standard)

womullan:


texmf/bibtex/bib/refs.bib
```diff
@@ -2344,4 +2344,58 @@ @article{e6805d07-4fd7-31d0-9779-42b323de329b
  volume = {49},
  year = {1984}
 }
-
+@article{kinnear1998,
```
timj: In the absence of ADS ID can we please use DOI as the key rather than author year? Author year can clash but DOI never will.

PR #2084, opened 2026-07-31
Remove the \c macro

timj:
The \c macro is part of tex and is used to write cedillas. This macro made it impossible to write some names without using unicode.


womullan (approved):
Yea seems odd

PR #2095, opened 2026-08-07
new LINEA address

womullan:


etc/authordb.yaml
```diff
@@ -1640,11 +1640,13 @@ affiliations:
     address:
       city: Rio de Janeiro
       country_code: Brazil
-      example_expanded: Laboratorio Interinstitucional de e-Astronomia, Rua General
-        Jose Cristino, 77, Rio de Janeiro, RJ, 20921-400, Brazil
-      postcode: 20921-400
+      example_expanded: Laboratorio Interinstitucional de e-Astronomia, i
```
gschwend: - [x] there is an extra ", i" in the end of the name (maybe a typo ?) 
- [x] if it is possible to include diacritical marks, there is a "ó" in "Laboratório" 

So the correct name is: 

example_expanded: Laboratório Interinstitucional de e-Astronomia

etc/authordb.yaml
```diff
@@ -1640,11 +1640,13 @@ affiliations:
...
+        Av. Pastor Martin Luther King Jr, 126 - Del Castilho Nova América Offices,
+        Torre 3000 / sala 817.
+        20765-000 – Rio de Janeiro - RJ, Brasil
+      postcode: 20765-000
       state: RJ
-      street: Rua General Jose Cristino, 77
+      street: Av. Pastor Martin Luther King Jr, 126 - Del Castilho Nova América Offices, Torre 3000 / sala 817.
     department: null
     email: linea.org.br
     institute: Laboratorio Interinstitucional de e-Astronomia
```
gschwend: "Laboratório" 

If it doesn't support diacritical marks, "Laboratorio" is ok.

gschwend (changes requested):
There is an extra ", i" in the end of the Lab's full name. 

Diacritical marks are optional.

etc/authordb.yaml
timj: You can use standard latex markup to get accents.

gschwend (approved):

Issue #2099, opened 2026-08-11
add new author


gschwend:
Please add Luigi Silva to the authors list so he can be added as co-author in RTN-124 after significant contribution to photo-z science unit of commissioning team. 

Following the standards in the Google Sheet "AuthorDB list":  

* Suggested Rubin AuthorID: `silval`
* Name: Luigi Lucas de Carvalho Silva 
* Affiliation ID(s): LINEA
* AASTEX: 
  - \author[0009-0009-5679-6347,gname='Luigi',sname='Silva']{Luigi-Silva}
  - \affiliation{Laborat\'orio Interinstitucional de e-Astronomia, Av. Pastor Martin Luther King Jr, 126 - Del Castilho Nova Am\'erica Offices, Torre 3000 / sala 817. 20765-000 – Rio de Janeiro - RJ, Brasil}
  - \email{luigi.silva AT linea.org.br}

timj:
Can you make a pull request to authordb.yaml?

The aastex formatting is not  relevant. It's much more structured. LINEA is already a defined affiliation so there is only a need to make the author entry.

gschwend:
@timj can you give permission to my user to push a new branch?

PR #2100, opened 2026-08-11
add new author

gschwend:
Adds Luigi Silva (LINEA) to the authors list file and closes issue #2099.

### DIFF
--- a/etc/authordb.yaml
+++ b/etc/authordb.yaml
@@ -8693,7 +8693,7 @@ authors:
     family_name: Silva
     given_name: Cristi\'{a}n
     orcid: 0009-0000-4228-4150
-  silval:
+  silvallc:
     affil:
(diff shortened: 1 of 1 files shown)


timj (approved):
Looks good. Are they okay with the shorter form of their name in papers?

gschwend:
@timj good point. I double checked with him and indeed, he prefers using all initials of middle names. I updated the file following the same pattern as other people with given name plus initials in a new commit today. Let me know if it is correct. Thanks!

timj:
Great. Please change the author ID to silvallc so that we can use those extra initials to ensure uniqueness if other Silvas turn up. Then I will merge.

timj (approved):

timj (approved):
Thanks. I will merge when checks complete.

PR #2117, opened 2026-08-27
Automatic authordb  update

womullan:


etc/authordb.yaml
```diff
@@ -808,6 +808,19 @@ affiliations:
...
     ror_id: 01sbq1a82
+  DoAUoM:
+    address:
+      city: ''
+      country_code: USA
+      example_expanded: Department of Astronomy, University of Michigan, 1085 South
+        University Avenue, Ann Arbor, MI 48109, USA
+      postcode: '48109'
+      state: null
+      street: University of Michigan, 1085 South University Avenue, Ann Arbor, MI
```
timj, line 819: This all needs fixing since the state and city are messed up.

timj (changes requested):

etc/authordb.yaml
```diff
@@ -808,6 +808,19 @@ affiliations:
     email: udel.edu
     institute: University of Delaware
     ror_id: 01sbq1a82
+  DoAUoM:
```
timj, line 811: I also don't like this affiliation starting with "DoA" for Department of Astronomy since we really want the UMich to be at the start so at least UMichDoA.

timj:
Superseded by #2121

PR #2121, opened 2026-08-28
Automatic authordb  update

womullan:


timj (changes requested):
Still just as broken as #2117 

I guess the idea is that these keep coming every day until I fix the PR because we have no means of requesting a fix from upstream...

PR #2127, opened 2026-09-01
Automatic bibfile  update

womullan:

### DIFF
--- a/texmf/bibtex/bib/lsst.bib
+++ b/texmf/bibtex/bib/lsst.bib
@@ -10955,15 +10955,15 @@
     url = "https://rtn-046.lsst.io/"
 }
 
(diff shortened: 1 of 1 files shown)


texmf/bibtex/bib/lsst.bib
```diff
@@ -10955,15 +10955,15 @@ @TechReport{RTN-046
     url = "https://rtn-046.lsst.io/"
 }
 
-@TechReport{RTN-047,
-    author = "Guy, Leanne",
-    title = "{System Performance Management Plan}",
+@TechReport{rtn-047,
```
timj, line 10958: This is odd. Why did it suddenly change to lower case key.

PR #2133, opened 2026-09-09
 use browse for bib file

womullan:


bin/generateBibfile.py
```diff
@@ -103,13 +98,15 @@ async def generate_bibfile(
     result : `str`
         Formatted bib file string ready to be printed.
     """
-    if not query:
-        query = ""  # Algolia take None as string literal None
+    client = SearchClientSync(app_id="0OJETYIVL5", api_key=os.environ["DOCS_API"])
```
timj: This is going to have to be something like:
```python
api_key = os.environ.get("DOCS_API")
if not api_key:
    raise RuntimeError("Unable to obtain algolia API key from DOCS_API environment variable"
```

womullan: yea i was being lazy ..

